PR #7, opened 2024-11-22
ALFMOB-33: Implement CI/CD using Github Actions

naveenmindera:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-33

### Description

- Added GitHub Actions and Fastlane workflow for CI/CD 
- Run tests code coverage on each pull request  
- Store test results (`.xcresult`) as GH Action artifacts  
- Distribute TestFlight builds on each pull request merged to `main`
- Manage app build number via git tags (#123 -> #124)


Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
+---
+name: Alfie CI/CD
+
+on:
+  push:
+    branches:
+      - main
+      - release/**
+      - ALFMOB-33_Implemented_CI_CD
```
p4checo, line 9: I don't think we should keep this, right? 😄

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+  workflow_dispatch:
+  schedule:
+    - cron: '0 23 * * 0' # Every Sunday at 23:00 GMT
```
p4checo, line 12: do we really need this, if we build for every merge to `main`? 🤔

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+
+      - name: Set up Xcode
+        uses: maxim-lobanov/setup-xcode@v2
+        with:
+          xcode-version: 16
+
+      - name: Install SwiftLint
+        run: brew install swiftlint
+
+      - name: Run Static Code Analysis
```
p4checo, line 29: this is not static code analysis, it's linting, right?

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: swiftlint
+
+      - name: Run Unit Tests
+        run: xcodebuild test -scheme Alfie -destination
+          'platform=iOS Simulator,name=iPhone 14'
```
p4checo, line 34: the destination should probably be an env var, so we can change it easily as needed

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Set up Xcode
+        uses: maxim-lobanov/setup-xcode@v2
+        with:
+          xcode-version: 16
+
+      - name: Increment Build Number
+        run: agvtool next-version -all
```
p4checo, line 49: where is the new build number stored? If this always increments from what's on the project then it will always generate the same BN, right?

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+  branch-validation:
+    runs-on: macos-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v3
+
+      - name: Set up Xcode
+        uses: maxim-lobanov/setup-xcode@v2
+        with:
+          xcode-version: 16
```
p4checo, line 24: The [`macos-latest`](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#xcode)(`macos-14-arm64`) image we're using doesn't appear to have Xcode 16 in it, but the [`macos-14`](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#xcode) (non arm) does, which is odd.

Anyway, I think we can pick the Xcode version by setting an env variable instead of using an action, e.g.
`DEVELOPER_DIR: '/Applications/Xcode_16.1.app/Contents/Developer'`

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,75 @@
...
+          api_key_id: ${{ secrets.API_KEY_ID }}
+          api_issuer: ${{ secrets.API_ISSUER }}
+          ipa_path: ${{ github.workspace }}/build/App.ipa
+
+  debug-build-schedule:
+    needs: branch-validation
+    runs-on: macos-latest
+    steps:
+      - name: Trigger Debug Build
+        run: echo "Scheduled Debug Build for TestFlight"
```
p4checo, line 75: this seems to be incomplete, yes?

p4checo (changes requested):
Nice one! Still some pending questions though 🤓

Alfie/.github/workflows/alfie.yml
reisdev: We could also use `xcodes` cli and a `.xcode-version` file, and leverage brew to do this setup.

Alfie/Alfie.xcodeproj/project.pbxproj
```diff
@@ -1015,6 +1017,22 @@
...
+			sourceTree = "<group>";
+		};
+		FCA0B27D2CEF3EFB000339F6 /* workflows */ = {
+			isa = PBXGroup;
+			children = (
+				FCA0B27E2CEF3F12000339F6 /* alfie.yml */,
+			);
+			path = workflows;
+			sourceTree = "<group>";
+		};
```
reisdev, line 1035: There's no need to add the `.github` folder to the `.xcodeproj`, neither the `alfie.yml` file

Alfie/.github/workflows/alfie.yml
naveenmindera: What I read about agvtool is, it increases the "CURRENT_PROJECT_VERSION" in project file, so the build number stores in CURRENT_PROJECT_VERSION. But I have to test this, once the workflow is ready.

Alfie/.github/workflows/alfie.yml
p4checo: why add even more tooling/setup when a good old env var is enough? 😅

Alfie/.github/workflows/alfie.yml
p4checo: Ok, maybe the question should be: How will build numbers be **managed**?, i.e:
1. when are they incremented, and by who?
3. how are they stored/persisted across builds?

Alfie/Alfie.xcodeproj/project.pbxproj
p4checo: I think the idea is that we can edit these files while working on the project in Xcode, instead of having to open another editor. 

To your point, I think we might not need to add the whole folder and just add the relevant files.

IMO we just need to be careful that these files aren't added to the app target, but I don't see any problem in adding them to the project as references. In other projects I've worked on we even create a group _without folder_, and add file references to it for all these CI/tooling/setup/documentation files

Alfie/.github/workflows/alfie.yml
naveenmindera: Build numbers are incremented by the pipeline whenever a new build is triggered, but this build number will exist temporarily in the pipeline. Now it makes sense, we need to commit back the version number to the repo, to maintain the update version across builds

Alfie/fastlane/Appfile
```diff
@@ -0,0 +1,8 @@
+app_identifier("com.mindera.alfie.debug") # The bundle identifier of your app
+apple_id("naveen.katari@mindera.com") # Your Apple Developer Portal username
```
p4checo, line 2: I don't think this is correct or needed. If we use API Key auth we don't need an Apple ID

Alfie/.github/workflows/alfie.yml
```diff
@@ -60,16 +97,5 @@ jobs:
...
-          ipa_path: ${{ github.workspace }}/build/App.ipa
-
-  debug-build-schedule:
-    needs: branch-validation
-    runs-on: macos-latest
-    steps:
-      - name: Trigger Debug Build
-        run: echo "Scheduled Debug Build for TestFlight"
+        run: |
+         fastlane pilot upload -i $PWD/build/App.ipa
```
p4checo, line 101: we're not using the `beta` lane anywhere, or am I missing something? 🤔

p4checo (changes requested):
We're on the right path, but still some rough edges here and there 😄

Alfie/fastlane/Fastfile
```diff
@@ -13,13 +13,41 @@
...
 
-default_platform(:ios)
-
 platform :ios do
-  desc "Push a new beta build to TestFlight"
+  desc "Validate branch with linting, analysis, and tests"
+  lane :validate_branch do
+    sh("swiftlint") # Runs SwiftLint
+    sh("xcodebuild analyze -scheme Alfie -destination '#{ENV['IOS_DESTINATION']}'")
+    sh("xcodebuild test -scheme Alfie -destination '#{ENV['IOS_DESTINATION']}'")
```
p4checo, line 21: perhaps use `run_tests` action (aka `scan`) ?

https://docs.fastlane.tools/getting-started/ios/running-tests/

We can also set `SCAN_DESTINATION` and any other env var once which then gets passed automatically to each fastlane action. Take a look [here](https://github.com/fastlane/fastlane/blob/master/scan/lib/scan/options.rb) for scan's parameters, for instance.

Alfie/fastlane/Fastfile
```diff
@@ -13,13 +13,41 @@
 # Uncomment the line if you want fastlane to automatically update itself
 # update_fastlane
 
-default_platform(:ios)
-
 platform :ios do
-  desc "Push a new beta build to TestFlight"
+  desc "Validate branch with linting, analysis, and tests"
+  lane :validate_branch do
+    sh("swiftlint") # Runs SwiftLint
```
p4checo, line 19: there's a [swiftlint](https://docs.fastlane.tools/actions/swiftlint/) fastlane action. I have mixed feelings about bundling `swiftlint` executable in the project, but I tend to prefer it being povided externally (e.g. via brew, SPM plugin, ...)

Alfie/fastlane/Fastfile
```diff
@@ -13,13 +13,41 @@
...
+    sh("xcodebuild analyze -scheme Alfie -destination '#{ENV['IOS_DESTINATION']}'")
+    sh("xcodebuild test -scheme Alfie -destination '#{ENV['IOS_DESTINATION']}'")
+  end
+
+  desc "Deploy to TestFlight using Git tags"
   lane :beta do
-    increment_build_number(xcodeproj: "Alfie.xcodeproj")
-    build_app(scheme: "Alfie")
-    upload_to_testflight
+    api_key = app_store_connect_api_key(
```
p4checo, line 26: I don't think we need to store the `api_key`, as this action stores it on the environment for subsequent actions to use (such as `upload_to_testflight`, `match`, etc)

Alfie/fastlane/Fastfile
```diff
@@ -13,13 +13,41 @@
...
+      key_id: ENV['ASC_KEY_ID'],
+      issuer_id: ENV['ASC_ISSUER_ID'],
+      key_filepath: ENV['ASC_API_KEY']
+    )
+
+    match(type: "appstore", readonly: false, api_key: api_key)
+
+    increment_build_number
```
p4checo, line 34: we are incrementing the build number, but how is it updated after each build?

Shouldn't we get the last tag's value, extract the BN,  then increment and set it on the Xcode project instead?

Alfie/fastlane/Fastfile
```diff
@@ -13,13 +13,41 @@
...
+
+    version = get_version_number(xcodeproj: "Alfie.xcodeproj")
```
p4checo, line 36: we can and probably should store the Xcode project's path on an env var, which avoids duplication and possible errors.

Alfie/.github/workflows/alfie.yml
p4checo: Continuing discussion on the new changes

p4checo (changes requested):
nice additions! We're on the right path 😉

Alfie/fastlane/Fastfile
JoaoPinhoMinder: Like we talked, added `swiftlint` and `xcbeautify` in a `Brewfile` here: [ALFMOB-33: Add environment variables file and Brewfile](https://github.com/Mindera/Alfie-iOS/pull/7/commits/4f56e0dd393fe559bbf048aa4dad8abbbd2ef8ec)

Alfie/fastlane/Fastfile
JoaoPinhoMinder: Updated like we talked on [ALFMOB-33: Adjust environment variables and update script](https://github.com/Mindera/Alfie-iOS/pull/7/commits/1d03213baae363ab449aaa560cc794e13a269ab3)

Alfie/fastlane/Fastfile
JoaoPinhoMinder: Project version isn't updated anymore like we talked. JFY, now we only update build number and tag it, app version is manually managed for now: [ALFMOB-33: Adjust environment variables and update script](https://github.com/Mindera/Alfie-iOS/pull/7/commits/1d03213baae363ab449aaa560cc794e13a269ab3)

JoaoPinhoMinder:
The original requirements have change a little, so will ask for a full review of this pull request.

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
+# This file contains the fastlane.tools configuration
```
p4checo, line 1: I think we can remove this header comment boilerplate

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+#
+#     https://docs.fastlane.tools/plugins/available-plugins
+#
+
+# Uncomment the line if you want fastlane to automatically update itself
+# update_fastlane
+
+platform :ios do
+  desc "Validate Tests"
+  lane :validate_tests do
```
p4checo, line 18: as mentioned previously, perhaps `test` does the trick (And "Run Tests" as description). To avoid collisions with `run_tests`, we can also use the actions' original names, `scan` instead of `run_tests`, `gym` instead of `build_app`, etc

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+
+    sh("brew bundle install --file=../Brewfile")
+
+    run_tests(
+      scheme: ENV['SCHEME']
+    )
+  end
+
+  desc "Deploy to TestFlight using Git tags"
```
p4checo, line 27: Perhaps "Build and deploy app release to TestFlight"?

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+  lane :beta do
```
p4checo, line 28: we're not calling it Beta, so perhaps just `release` does the trick

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+    new_build_number = last_build_number + 1
+
+    UI.message("🔢 Last Build Number: #{last_build_number}")
+    UI.message("🚀 New Build Number: #{new_build_number}")
+
+    increment_build_number(build_number: new_build_number)
+
+    new_tag = "'##{new_build_number}'"
+
+    sh("git config user.name '#{ENV['GIT_USER_NAME']}'")
```
p4checo, line 49: to make our logs a bit less verbose, perhaps we could bundle these shell commands into just two (e.g. with `;` between commands). One to set user name and email, and another to set the tag and push 💅🏼

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+      build_configurations: [ENV['GYM_CONFIGURATION']],
+      code_sign_identity: signing_identity(type: "appstore"),
+      profile_name: provisioning_profile_name(type: "appstore")
+    )
+
+    UI.message("#️⃣ Fetching Build Number from latest tag...")
+    last_build_number = sh("git tag -l '#*' --sort=-v:refname | head -n 1 | sed -e 's/#//g' | grep . || echo '0'").strip.to_i
+    new_build_number = last_build_number + 1
+
+    UI.message("🔢 Last Build Number: #{last_build_number}")
```
p4checo, line 42: we could merge these 2 log messages into a single one, like: `🔢 New Build Number: #{last_build_number} -> #{new_build_number}")`

The `increment_build_number` also logs the BN that it sets I think.

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+
+    new_tag = "'##{new_build_number}'"
+
+    sh("git config user.name '#{ENV['GIT_USER_NAME']}'")
+    sh("git config user.email '#{ENV['GIT_EMAIL']}'")
+    sh("git tag #{new_tag}")
+    sh("git push origin #{new_tag}")
+
+    build_app(
+      scheme: ENV['SCHEME'],
```
p4checo, line 55: this can be set via `GYM_SCHEME` in `.env.default`, similar for the other parameters.

Please check: https://github.com/fastlane/fastlane/blob/master/gym/lib/gym/options.rb

Alfie/Brewfile
```diff
@@ -0,0 +1 @@
+brew "xcbeautify"
```
p4checo, line 1: please add a newline at the end of each file (multiple cases)

Alfie/fastlane/.env.default
```diff
@@ -0,0 +1,86 @@
...
+# Global
+
+ROOT_PATH="./"
+PROJECT_PATH="${ROOT_PATH}/Alfie.xcodeproj"
+
+SCHEME="Alfie"
+TESTPLAN="Alfie"
+IOS_DESTINATION="platform=iOS Simulator,name=iPhone 16"
+
+BUILDLOGS_PATH=${ARTIFACTS_PATH}
```
p4checo, line 11: `ARTIFACTS_PATH` is not being defined

Alfie/fastlane/.env.default
```diff
@@ -0,0 +1,86 @@
...
+
+TEAM_ID=${APPSTORE_CONNECT_TEAM_ID}
+
+BUILD_CONFIGURATION="Release"
```
p4checo, line 15: This should only be set to `Release` for release builds

Alfie/fastlane/.env.default
```diff
@@ -0,0 +1,86 @@
...
+GYM_CLEAN=true
+
+GYM_INCLUDE_BITCODE=false
+GYM_EXPORT_OPTIONS="fastlane/gym_exportOptions.plist"
+
+GYM_BUILDLOG_PATH=${BUILDLOGS_PATH}
+
+GYM_SDK="iphoneos"
+
+GYM_CLONED_SOURCE_PACKAGES_PATH=${SPM_CLONED_DEPENDENCIES_PATH}
```
p4checo, line 73: `SPM_CLONED_DEPENDENCIES_PATH` is not being defined anywhere. We typically use this to control where SPM  dependencies are cloned so that we can cache them and/or access any binaries within them (e.g. Crashlytics symbol upload script)

That reminds me that we should migrate Crashlytics symbol upload to be made only in CI release builds, not on every build via a build phase (as it is now) 😅

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Validate Tests
+        run: fastlane validate_tests
+        env:
+          IOS_DESTINATION: ${{ env.IOS_DESTINATION }}
+      
+      - name: Archive Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: iOS-Test-Results
```
p4checo, line 35: perhaps `Test artifacts`?

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,71 @@
...
+    branches:
+      - 'main'
+
+  workflow_dispatch:
+
+env:
+  DEVELOPER_DIR: "/Applications/Xcode_16.app/Contents/Developer"
+
+jobs:
+  branch-validation:
```
p4checo, line 18: not a big fan of this name, perhaps `test`? (Don't forget to update the `needs:` clause in the release job)

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,71 @@
...
+      
+      - name: Archive Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: iOS-Test-Results
+          path: |
+              ~/Library/Developer/Xcode/DerivedData/**/Logs/Test/*.xcresult
+          retention-days: 7
+
+  testflight-build:
```
p4checo, line 40: perhaps `release`?

Alfie/.github/workflows/alfie.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          mkdir -p ~/.ssh
+          echo "${{ secrets.BITBUCKET_SSH_KEY }}" > ~/.ssh/id_rsa
+          chmod 600 ~/.ssh/id_rsa
+          ssh-keyscan -t rsa bitbucket.org >> ~/.ssh/known_hosts
+
+      - name: Install Fastlane
+        run: gem install fastlane
+
+      - name: Deploy to TestFlight
```
p4checo, line 58: "Build and deploy release to TestFlight" 💅🏼

Alfie/fastlane/Fastfile
```diff
@@ -0,0 +1,100 @@
...
+      export_options: { method: "app-store" },
+      configuration: ENV['BUILD_CONFIGURATION']
+    )
+
+    app_store_connect_api_key
+
+    upload_to_testflight
+  end
+
+  # Helper lane to renew Code Signing certificates/provisioning profiles
```
p4checo, line 65: as detailed on the code signing document,  please add a `mock-ci-env.sh` script in a `scripts` folder so it's easier to mock the CI environment locally in the future. Be careful not to commit any sensitive data though 😆

p4checo (changes requested):
Great work! 💪🏼 
Still some things that need improvement and cleaning up 😉

Alfie/.github/workflows/alfie.yml
```diff
@@ -12,34 +12,34 @@ on:
   workflow_dispatch:
 
 env:
-  DEVELOPER_DIR: "/Applications/Xcode_16.app/Contents/Developer"
+  DEVELOPER_DIR: "/Applications/Xcode_16.2.app/Contents/Developer"
+  ARTIFACTS_PATH: "~/Library/Developer/Xcode/DerivedData/Logs/Test"
```
p4checo, line 16: we don't need to store it on the DerivedData folder, something like `/tmp/alfie-artifacts` also works fine 💅🏼

Alfie/Gemfile
```diff
@@ -0,0 +1,3 @@
+source "https://rubygems.org"
+
+gem "fastlane"
```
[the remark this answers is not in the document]
p4checo: that's running locally, right? 😆 I assume that you're invoking it as `bundle exec fastlane <lane>` locally which uses bundler, however in CI we're just invoking `fastlane <lane>`

Nonetheless, I think that we should probably add a step to install Homebrew formulas and another to install Ruby dependencies, so in case the GH image stops having these tools our CI will keep working. It also prepares the setup in case we add any Homebrew/Ruby dependency that's not bundled in the GH image by default.

Alfie/Brewfile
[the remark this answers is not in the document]
p4checo: You're right. Missed that 🙈 

However, we don't typically manage/install dependencies from fastlane, but rather as part of CI (as in steps). Please see my other comment regarding Homebrew dependencies 😉

p4checo (changes requested):
looking good! We're almost there 😉

Alfie/.github/workflows/alfie.yml
```diff
@@ -13,8 +13,9 @@ on:
 
 env:
   DEVELOPER_DIR: "/Applications/Xcode_16.2.app/Contents/Developer"
-  ARTIFACTS_PATH: "~/Library/Developer/Xcode/DerivedData/Logs/Test"
+  ARTIFACTS_PATH: "/tmp/alfie-artifacts"
   DESTINATION: "platform=iOS Simulator,OS=latest,name=iPhone 16 Pro" 
+  SPM_CLONED_DEPENDENCIES_PATH: "tmp/SourcePackages"
```
p4checo, line 18: as discussed, please use `/tmp/foo` and not `tmp/foo` as the first is an absolute path (on the machine's root) and the second is a relative path (to the project folder)

p4checo (changes requested):
small fixes

p4checo (approved):
🚀

reisdev (approved):

PR #8, opened 2024-11-22
[ALFMOB-24] Feature Toggle on debug menu

reisdev:
## Changelog

* Create `FeatureToggle` screen
* Change `LocalConfigurationProvider` to use `UserDefaults` instead of a json file
* Add wildcard to `.gitignore` to avoid adding every new SPM package

## Evidence

| Debug config enabled | Debug config disabled |
|:----------------------:|:-----------------------|
| Simulator Screenshot - iPhone 16 Pro - 2024-11-22 at 14 52 55 | Simulator Screenshot - iPhone 16 Pro - 2024-11-22 at 14 52 59 |

https://github.com/user-attachments/assets/4cae7124-fefd-4b76-9916-b88804947fc3


Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleView.swift
```diff
@@ -0,0 +1,36 @@
...
+
+struct FeatureToggleView<ViewModel: FeatureToggleViewModelProtocol>: View {
+	@ObservedObject private var viewModel: ViewModel
+
+	init(viewModel: ViewModel) {
+		self.viewModel = viewModel
+	}
+
+	var body: some View {
+		LazyVGrid(columns: [GridItem()]) {
```
p4checo, line 13: why a `LazyVGrid` and not a `LazyVStack`? 🤔

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleView.swift
```diff
@@ -0,0 +1,36 @@
...
+			ForEach(viewModel.features.sorted { $0.key > $1.key }, id: \.key) { feature, isEnabled in
```
p4checo, line 14: this business logic should be on the VM

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,39 @@
+import FeatureToggles
+import Models
+import SwiftUI
+
+final class FeatureToggleViewModel: FeatureToggleViewModelProtocol {
+	@Published private(set) var features: [String: Bool] = [:]
+
+	private let service: any ConfigurationServiceProtocol
+
+	init(service: some ConfigurationServiceProtocol) {
```
p4checo, line 10: why do we need `some` here? 🤔 

also, shouldn't this receive the provider it's attached to, instead of the whole service?

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,39 @@
...
+	func viewDidAppear() {
+		let featuresAsDict = ConfigurationKey.allCases.map { configuration in
+			(configuration.rawValue, service.isFeatureEnabled(configuration))
+		}
+
+		self.features = Dictionary(uniqueKeysWithValues: featuresAsDict)
+	}
+
+	func didUpdate(feature: String) {
+		guard
```
p4checo, line 23: if the `guard` fits a single line, we don't need to break it

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,39 @@
+import FeatureToggles
+import Models
+import SwiftUI
+
+final class FeatureToggleViewModel: FeatureToggleViewModelProtocol {
+	@Published private(set) var features: [String: Bool] = [:]
```
p4checo, line 6: couldn't we use `[ConfigurationKey: Bool]` instead?

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+struct LocalizableFeatureToggle: LocalizableProtocol {
+	@LocalizableResource<Self>(.title) static var title
+
+	static func featureName(for feature: String, locale: Locale = .current) -> LocalizedStringResource {
+		let identifier = feature.capitalized.replacingOccurrences(of: "_", with: "")
```
p4checo, line 7: perhaps using the config key's raw value is better and less error prone? Also not sure about having to use `Key<>Feature`

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
```diff
@@ -115,7 +119,10 @@ extension ConfigurationService {
     }
 
     private func checkFeatureAvailability(key: ConfigurationKey) -> Bool {
-        guard let rawValue = providerValue(for: key), let configValue = ConfigurationValue(rawValue: rawValue) else {
+        guard
```
p4checo, line 122: no need to break this `guard` if it fits 120 chars (multiple cases)

Alfie/Packages/Models/Sources/Services/Configuration/ConfigurationProviderProtocol.swift
```diff
@@ -10,4 +10,10 @@ public protocol ConfigurationProviderProtocol {
     func double(for key: ConfigurationKey) -> Double?
     func int(for key: ConfigurationKey) -> Int?
     func string(for key: ConfigurationKey) -> String?
+
+	func set(_ value: Bool, for key: ConfigurationKey)
```
p4checo, line 14: we should not change the `ConfigurationProviderProtocol` API to allow writing configs. This should be a capability of a single provider

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
```diff
@@ -76,6 +76,10 @@ public final class ConfigurationService: ConfigurationServiceProtocol {
         return value
     }
 
+	public func updateFeature(_ key: ConfigurationKey, isEnabled: Bool) {
```
p4checo, line 79: the `ConfigurationService` should be read only

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		let isEnabled = service.isFeatureEnabled(key)
+
+		service.updateFeature(key, isEnabled: !isEnabled)
+
+		self.features[feature] = service.isFeatureEnabled(key)
+	}
+
+	func description(for feature: String) -> LocalizedStringResource {
```
p4checo, line 36: if we use `ConfigurationKey` then we can add a computed property for the localized text

p4checo (changes requested):
Nice work! Some suggestions and a bit of clean up missing 😄

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleView.swift
reisdev: I have the same question? 😅

Consider it fixed.

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
reisdev: I decided to pre-process it because, otherwise, we'll have localisation keys like `Keyios_wishlistFeature`, which seemed out-of-place.

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
reisdev: Ok

Alfie/Packages/Models/Sources/Services/Configuration/ConfigurationProviderProtocol.swift
reisdev: Ok, I'll undo it.

Alfie/Alfie/Service/ServiceProvider.swift
```diff
@@ -39,8 +39,15 @@ final class ServiceProvider: ServiceProviderProtocol {
             minimumFetchInterval: ReleaseConfigurator.isDebug ? 30 : 1800
         )
         let localProvider = LocalConfigurationProvider()
+
+		#if DEBUG
+		let providers: [ConfigurationProviderProtocol] = [localProvider]
```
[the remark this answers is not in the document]
reisdev: Ok, I can get it to work this way. The only problem that I see is:

> How to bypass all the providers and use only the `DebugConfigurationProvider`?

Any ideas?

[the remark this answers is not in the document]
p4checo: if the `DebugConfigurationProvider` is the first of the chain, then if it returns a value (and not `nil`) it will prevail over all other providers. That's how the pattern being used (chain of responsibility) works.

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
p4checo: Not related to your changes, but I never understood the `KeyFooBar` approach. Personally I always used sort of reverse domain notation with snake_case, and I think it's better for most cases and this is no exception 😅 

So I would see this as `feature_toggle.<feature_toggle_name>`, or even the `<feature_toggle_name>` directly. However, to not go against the current localization _status quo_, perhaps we can leave it as is.

Alfie/Alfie/Service/ServiceProvider.swift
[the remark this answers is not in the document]
reisdev: So the toggle for a "global override" would force the provider to return `nil` when it's not enabled, right?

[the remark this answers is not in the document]
p4checo: Precisely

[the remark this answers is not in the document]
timea-v: Wouldn't this approach cause too many user interaction to get to a single feature? Maybe we could just have set remote toggle by default in debug menu for all features and only if that changes to on/off we set the debug config value in `DebugConfigurationProvider`, otherwise it would be `nil`?

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
reisdev: I agree with you and prefer the same pattern, but using camel-case. I've just followed the current localisation keys pattern.

.gitignore
```diff
@@ -120,11 +120,5 @@ fastlane/screenshots/**/*.png
...
-StyleGuide/.swiftpm/xcode/xcuserdata/*
-Alfie/Alfie.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
-Alfie/Packages/Core/.swiftpm/xcode/xcuserdata/*
-Alfie/Packages/Models/.swiftpm/xcode/xcuserdata/*
-Alfie/Packages/Mocks/.swiftpm/xcode/xcuserdata/*
-Alfie/Packages/Navigation/.swiftpm/xcode/xcuserdata/*
-Alfie/Packages/StyleGuide/Package.resolved
-Alfie/Packages/Common/.swiftpm/xcode/xcuserdata/*
+**/*/.swiftpm
+**/*/swiftpm
```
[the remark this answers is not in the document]
reisdev: `Package.resolved` is intended, mostly, for local use. Even the SPM repo ignores this file:

https://github.com/swiftlang/swift-package-manager/blob/main/.gitignore

Alfie/Alfie/Delegate/AppDelegate.swift
```diff
@@ -54,7 +54,15 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
                 serviceProvider.notificationsService.start()
             }
         }
-        tabCoordinator = TabCoordinator(tabs: TabScreen.allCases, activeTab: .home(), serviceProvider: serviceProvider)
+
+        var tabs: [TabScreen] = TabScreen.allCases
+
+        // TODO: Review feature toggle flow
+        if !serviceProvider.configurationService.isFeatureEnabled(.wishlist) {
```
timea-v, line 61: Maybe we could have `var tabs` with all cases but `.wishlist` and just append `.wishlist` if it is enabled for better performance?

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    private let provider: any DebugConfigurationProviderProtocol
+    private var store: Set<AnyCancellable> = []
+
+    init(provider: some DebugConfigurationProviderProtocol) {
+        self.provider = provider
+    }
+
+    func viewDidAppear() {
+        self.features = ConfigurationKey.allCases.map { configuration in
```
timea-v, line 17:
```suggestion
        features = ConfigurationKey.allCases.map { configuration in
```

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,51 @@
+import Combine
+import Models
+import SwiftUI
+
+final class FeatureToggleViewModel: FeatureToggleViewModelProtocol {
+    @Published private(set) var features: [(feature: String, isEnabled: Bool)] = []
+    @Published var isDebugConfigurationEnabled = false
+
+    private let provider: any DebugConfigurationProviderProtocol
+    private var store: Set<AnyCancellable> = []
```
timea-v, line 10: Maybe it could be a single subscription?
```suggestion
    private var subscription: AnyCancellable?
```

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+struct LocalizableFeatureToggle: LocalizableProtocol {
+	@LocalizableResource<Self>(.title) static var title
```
timea-v, line 4: Indentation is wrong

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Localisation/LocalizableFeatureToggle.swift
```diff
@@ -0,0 +1,15 @@
...
+    @LocalizableResource<Self>(.debugConfigurationEnabled) static var debugConfigurationEnabled
+    @LocalizableResource<Self>(.wishlistFeature) static var wishlistFeature
+    @LocalizableResource<Self>(.appUpdateFeature) static var appUpdateFeature
+
+	enum Keys: String, LocalizableKeyProtocol {
+		case title = "KeyFeatureToggle"
```
timea-v, line 10: Indentation is wrong

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
```diff
@@ -135,26 +135,23 @@ extension ConfigurationService {
...
-                    return appUpdate.requirements.immediate
-                case .flexible:
-                    return appUpdate.requirements.flexible
-                }
-                // swiftlint:enable vertical_whitespace_between_cases
-            }()
-        else {
-            return nil
-        }
+            let configuration: ConfigurationAppUpdateInfo = switch type {
```
timea-v, line 138: Can you please explain why this change? The rule was disabled to follow Mindera style guide

Alfie/Packages/Core/Sources/Services/Configuration/Providers/DebugConfigurationProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    private lazy var decoder = JSONDecoder()
+    private lazy var encoder = JSONEncoder()
+
+    private init() {
+        loadConfig()
+    }
+
+    // MARK: - ConfigurationProviderProtocol
+    public func toggleAvailability() {
```
timea-v, line 24: Missing new line

Alfie/Packages/Core/Sources/Services/Configuration/Providers/DebugConfigurationProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+    public var isReadyPublisher: AnyPublisher<Bool, Never> { isReadySubject.eraseToAnyPublisher() }
+
+    private lazy var decoder = JSONDecoder()
+    private lazy var encoder = JSONEncoder()
+
+    private init() {
+        loadConfig()
+    }
+
+    // MARK: - ConfigurationProviderProtocol
```
timea-v, line 23: Missing new line

Alfie/Packages/Core/Sources/Services/Configuration/Providers/DebugConfigurationProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public func updateFeature(_ key: ConfigurationKey, isEnabled: Bool) {
+        setValue(isEnabled, for: key)
+    }
+
+    // MARK: - Private
+
+    private func loadConfig() {
+        guard
```
timea-v, line 79: Could you please use this style for `guard` with multiple conditions in this PR for consistency and to follow Mindera style guide?

Alfie/Packages/Core/Sources/Services/Configuration/Providers/DebugConfigurationProvider.swift
```diff
@@ -0,0 +1,106 @@
...
+            let data = UserDefaults.standard.data(forKey: Self.userDefaultsKey),
+            let config = try? decoder.decode([String: Data].self, from: data)
+        else {
+            return
+        }
+
+        self.isReadySubject.value = UserDefaults.standard.bool(forKey: Self.debugConfigurationKey)
+        self.localConfig = config
```
timea-v, line 87:
```suggestion
        isReadySubject.value = UserDefaults.standard.bool(forKey: Self.debugConfigurationKey)
        localConfig = config
```

Alfie/Packages/Models/Sources/Features/FeatureToggleViewModelProtocol.swift
```diff
@@ -0,0 +1,11 @@
+import SwiftUI
+
+public protocol FeatureToggleViewModelProtocol: ObservableObject {
+    var features: [(feature: String, isEnabled: Bool)] { get }
+    var isDebugConfigurationEnabled: Bool { get }
+
+	func viewDidAppear()
+	func didUpdate(feature: String)
+	func localizedName(for feature: String) -> LocalizedStringResource
```
timea-v, line 9: Indentation is wrong

Alfie/Packages/Models/Sources/Services/Configuration/Providers/DebugConfigurationProviderProtocol.swift
```diff
@@ -0,0 +1,7 @@
+import Combine
```
timea-v, line 1: Is importing `Combine` needed?

Alfie/Alfie/Service/ServiceProvider.swift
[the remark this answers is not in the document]
timea-v: I meant too many toggles, it could be a single picker for each feature (remote/on/off), but not so important.

Alfie/Alfie/Views/DebugMenu/DebugMenuView.swift
```diff
@@ -92,6 +93,18 @@ struct DebugMenuView: View {
             .modifier(
                 ContainerDemoViewModifier(headerTitle: "Environment", embedInScrollView: false)
             )
+
+        case .featureToggle:
+            FeatureToggleView(
+                viewModel: FeatureToggleViewModel(
```
timea-v, line 99: Line wrapping is not needed here.

Alfie/Alfie/Views/DebugMenu/DebugMenuView.swift
```diff
@@ -92,6 +93,18 @@ struct DebugMenuView: View {
...
+                    provider: DebugConfigurationProvider.shared
+                )
+            ).modifier(
+                ContainerDemoViewModifier(
```
timea-v, line 103: Maybe not needed to wrap lines here also

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    init(provider: some DebugConfigurationProviderProtocol) {
```
timea-v, line 12:
```suggestion
    init(provider: DebugConfigurationProviderProtocol) {
```

Alfie/Packages/Models/Sources/Features/FeatureToggleViewModelProtocol.swift
```diff
@@ -0,0 +1,11 @@
...
+    func toggleDebugConfiguration()
+}
```
naveenmindera, line 11: I see the feature toggle has heavier dependency on FeatureToggleViewModelProtocol. Consider breaking this into smaller protocols (ISP).

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleView.swift
```diff
@@ -0,0 +1,43 @@
...
+                    .font(Font(theme.font.paragraph.bold.withSize(20)))
+            }
+
+            Divider()
+            Spacer(minLength: Spacing.space200)
+
+            if viewModel.isDebugConfigurationEnabled {
+                ForEach(viewModel.features, id: \.feature) { feature, isEnabled in
+                    ThemedToggleView(isOn: Binding(
+                        get: { isEnabled },
```
naveenmindera, line 30: Can we move this logic to viewModel to generate the bindings for each feature. A func that returns Binding<Bool> 

My idea is this can be as simple as 
ThemedToggleView(
                        isOn: viewModel.binding(for: feature)
                    )
thoughts?

Alfie/Alfie/Delegate/AppDelegate.swift
reisdev: That was one of my options, but this will imply on inserting the `.wishlist` at a specific position and that could lead to other problems if the items on the tab bar changes. Any way, it looks cleaner to insert if available rather then removing if it's not.

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
reisdev: It was a syntax improvement, to use switch assignment instead of a closure. I reverted the rule enable/disable removal.

Alfie/Packages/Core/Sources/Services/Configuration/Providers/DebugConfigurationProvider.swift
reisdev: Where can I find Mindera's style guide? I'm not used to it

Alfie/Packages/Models/Sources/Services/Configuration/Providers/DebugConfigurationProviderProtocol.swift
reisdev: Not anymore. Removed it.

Alfie/Alfie/Views/DebugMenu/FeatureToggle/FeatureToggleView.swift
reisdev: Great catch. I did some improvements and made the VM responsible for this.

Alfie/Packages/Models/Sources/Features/FeatureToggleViewModelProtocol.swift
reisdev: I was able to simplify this. Do you still see the need for multiple protocols?

Alfie/Alfie/Delegate/AppDelegate.swift
```diff
@@ -54,7 +54,14 @@ class AppDelegate: NSObject, UIApplicationDelegate, UNUserNotificationCenterDele
                 serviceProvider.notificationsService.start()
             }
         }
-        tabCoordinator = TabCoordinator(tabs: TabScreen.allCases, activeTab: .home(), serviceProvider: serviceProvider)
+
+        var tabs: [TabScreen] = [.home(nil), .shop(tab: nil), .bag]
```
timea-v, line 58: Home and shop cases have default values for parameters, you can skip adding them:
```suggestion
        var tabs: [TabScreen] = [.home(), .shop(), .bag]
```

Alfie/Packages/Core/Sources/Services/Configuration/ConfigurationService.swift
```diff
@@ -132,29 +132,25 @@ extension ConfigurationService {
...
-                    return appUpdate.requirements.immediate
-                case .flexible:
-                    return appUpdate.requirements.flexible
-                }
-                // swiftlint:enable vertical_whitespace_between_cases
-            }()
-        else {
-            return nil
-        }
+        guard let rawValue = providerValue(for: .appUpdate),
```
timea-v, line 135: Please keep each condition on new line
```suggestion
        guard 
            let rawValue = providerValue(for: .appUpdate),
```

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Extensions/ConfigurationKeyExtension.swift
```diff
@@ -0,0 +1,17 @@
+import Models
+import SwiftUI
+
+extension ConfigurationKey {
+    var localizedName: LocalizedStringResource {
+        switch self {
+        case .appUpdate:
+            LocalizableFeatureToggle.appUpdateFeature
+
```
timea-v, line 9: New line is not needed here:
"Add a newline between each case block in switch statements if at least one spans more than one line."

Alfie/Alfie/Views/DebugMenu/FeatureToggle/Extensions/ConfigurationKeyExtension.swift
```diff
@@ -0,0 +1,17 @@
...
+        case .wishlist:
+            LocalizableFeatureToggle.wishlistFeature
+
+        default:
```
timea-v, line 13: Maybe it would be better to provide value for each case instead of using default, so that in the future if a new `ConfigurationKey` is added, a compiler error will force us to provide correct localization for that.

timea-v: I meant to provide all the possible cases, here as I saw "case .custom(...):" is missing

reisdev: Sorry. Added the case

p4checo (changes requested):
Still some pending bits here and there 😇

reisdev:
@p4checothe style guide doesn't state that I should only break the `guard` clause into multiple line if it exceeds 120. What rule should I follow?

https://github.com/Mindera/swift-style-guide?tab=readme-ov-file#golden-path

p4checo:
@reisdev You are correct. I've edited my comment and resolved the conversation 😉 

Our styleguide is long overdue a revision, but until that happens we should honor it. Thank you 🙏🏼

naveenmindera (approved):

timea-v (approved):

.gitignore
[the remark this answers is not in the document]
p4checo: Maybe my explanation wasn't clear. We're **not supposed** to ignore the `.resolved` file inside the project, because Xcode uses it to store the project's / workspace's resolved dependencies 😅

p4checo (changes requested):

.gitignore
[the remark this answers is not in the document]
reisdev: Since it was already ignored, I assumed it should stay this way 😅. Now, I removed it.

p4checo (approved):

PR #10, opened 2024-11-26
ALFMOB-42: Use local mock server

timea-v:
Added local mock server references and updated README.md accordingly.


reisdev (approved):

Alfie/Alfie/Configuration/LinkConfiguration.swift
```diff
@@ -7,7 +7,7 @@ struct LinkConfiguration: LinkConfigurationProtocol {
     init(
         schemes: [String] = ["https", "http", "alfie"],
         hosts: [String] = [
-            "mock-server-rose.vercel.app",
+            "localhost:4000",
```
p4checo, line 10: should/could we place this in a single place, where it would be easier to update in the future? It seems we have the same endpoint/URL duplicated in many places. 💅🏼

p4checo (approved):
LGTM. Just a minor improvement suggestion 🤓

naveenmindera (approved):

PR #14, opened 2024-12-05
ALFMOB-49: Wishlist feature adjustments

timea-v:
Implemented Wishlist feature for the following scenarios:
- Shop, Bag screen right toolbar icon
- Account list item
- Changed display of wishlist in tab bar in real time (without needing to restart the app)

https://github.com/user-attachments/assets/3c26e8cb-88be-409e-8a45-3b973fa6438b


Alfie/Alfie/Modifiers/DefaultToolbarModifier.swift
```diff
@@ -13,9 +14,11 @@ extension View {
 struct DefaultToolbarModifier: ViewModifier {
     @EnvironmentObject var coordinator: Coordinator
     private let screen: Screen
+    private let viewModel: DefaultToolbarModifierViewModelProtocol
 
-    public init(screen: Screen) {
+    public init(screen: Screen, viewModel: DefaultToolbarModifierViewModelProtocol) {
```
p4checo, line 19: It's a bit strange to pass a VM into a modifier... 🤔  Wouldn't injecting a `isWishlistEnabled: Bool` parameter/property be enough? Or maybe even use the Coordinator since we already use to see if the Debug Menu is being presented?

This seems to have added significant complexity (more protocols, more conformances, more mocks) without much benefit.

p4checo (changes requested):

p4checo (approved):
🚀

Alfie/Alfie/Modifiers/DefaultToolbarModifier.swift
timea-v: I refactored it and injected the value on `Coordinatior`.  Please check my latest commits.

naveenmindera (approved):

PR #29, opened 2025-02-10
ALFMOB-59: Refactor Reachability

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-59

### Description

Refactor ReachabilityService to consolidate monitoring logic

- Merged ReachabilityMonitor into ReachabilityService, reducing complexity
- Introduced a dedicated DispatchQueue with `.utility` QoS for better performance
- Optimized network status updates to prevent unnecessary state changes

This refactor improves maintainability and enhances performance while preserving expected behavior.


Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
```diff
@@ -1,20 +1,35 @@
...
 import Models
+import Network
 
 public final class ReachabilityService: ReachabilityServiceProtocol {
-    @Published var isAvailable = false
+    @Published private var isAvailable = false
     public var networkAvailability: AnyPublisher<Bool, Never> { $isAvailable.eraseToAnyPublisher() }
     public var isNetworkAvailable: Bool { isAvailable }
 
-    private var monitor: ReachabilityMonitorProtocol
```
reisdev, line 9: Regarding removing the protocol and not injecting the monitor into this class: How would you test the start/stop action handling? How to validate if the method calls are reflected on the monitor?

JoaoPinhoMinder: The main goal of this refactor was to simplify the reachability logic by merging `ReachabilityMonitor` into `ReachabilityService`, reducing unnecessary indirection.

The primary behavior we care about is whether `isAvailable` correctly updates when network conditions change. In my opinion, mocking `NWPathMonitor` to validate if `startMonitoring` and `stopMonitoring` are called, adds minimal value.

This way we are focusing on testing observable behavior rather than internal methods calls.

Either way, i can add the `startMonitoring` and `stopMonitoring` to the `ReachabilityServiceProtocol` and do the tests or introduce the `ReachabilityMonitorProtocol` again and inject a mock monitor, however, i believe the main goal was to drop this last one 😄

Let me know your thoughts on this 🙏

reisdev: No need for the protocol itself, indeed. We could just enable inject a `NWPathMonitor` to enable more flexibility and testability.

p4checo: You raise valid points, and sadly `NWPathMonitor` is a `final class` so we can't subclass it to overload and mock it. 

We can however create a `NetworkPathMonitorProtocol` that mirrors the `NWPathMonitor`'s APIs we use so we can inject that instead (and add conformance to `NWPathMonitor`).

Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
```diff
@@ -1,20 +1,35 @@
...
+    private let monitor = NWPathMonitor()
+    private let utilityQueue = DispatchQueue(label: "com.mindera.alfiekit", qos: .utility)
```
p4checo, line 12: we should inject the queue as well, as we might want to use an existing queue or control it during tests.

Using reverse domain in the label is correct, but we should probably specify it's for reachability, so `"com.mindera.alfiekit.reachability"`

Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
```diff
@@ -1,20 +1,35 @@
...
-        self.monitor = monitor
-        monitor.networkAvailability.assign(to: &$isAvailable)
-        monitor.startMonitoring()
+    public init() {
+        startMonitoring()
+    }
+
+    private func startMonitoring() {
+        monitor.pathUpdateHandler = { [weak self] path in
+            let newStatus = path.status == .satisfied
```
p4checo, line 24: perhaps `guard let self else { return }` somewhere? it doesn't make sense to keep working if the service is dealloc'ed. right? 😅

Alfie/AlfieKit/Tests/CoreTests/ServiceTests/ReachabilityServiceTests.swift
```diff
@@ -3,72 +3,47 @@ import XCTest
...
 
     override func tearDownWithError() throws {
-        sut = nil
-        mockReachabilityMonitor = nil
-        timeout = nil
+        mockReachabilityService = nil
         try super.tearDownWithError()
     }
 
-    func test_reachabilityService_startsMonitorOnInit() {
```
p4checo, line 24: we're losing some test coverage because we're not injecting the monitor. Please review implementation to improve DI as per previous comments 🙏🏼

Alfie/AlfieKit/Tests/CoreTests/ServiceTests/ReachabilityServiceTests.swift
```diff
@@ -3,72 +3,47 @@ import XCTest
 @testable import Core
 
 final class ReachabilityServiceTests: XCTestCase {
-    private var sut: ReachabilityService!
-    private var mockReachabilityMonitor: MockReachabilityMonitor!
-    private var timeout: CGFloat!
+    private var mockReachabilityService: MockReachabilityService!
```
p4checo, line 6: it doesn't make sense to test a `Foo` implementation with a `MockFoo` instance, right? 🙃

Alfie/AlfieKit/Sources/TestUtils/Helpers/XCTestCase+Combine.swift
```diff
@@ -39,6 +39,7 @@ extension XCTestCase {
         wait(for: [expectation], timeout: timeout)
     }
 
+    @discardableResult
```
p4checo, line 42: nice improvement! 🙏🏼

p4checo (changes requested):
Nice one! We should do some additional improvements especially around testability 🙏🏼

Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
JoaoPinhoMinder: Updated, let me know if i meet the expected goal 🙏

Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
JoaoPinhoMinder: ✅

Alfie/AlfieKit/Sources/Core/Services/Reachability/ReachabilityService.swift
JoaoPinhoMinder: Since the implementation changed, there was no need to capture `self`.

Alfie/AlfieKit/Tests/CoreTests/ServiceTests/ReachabilityServiceTests.swift
JoaoPinhoMinder: Updated ✅

Alfie/AlfieKit/Tests/CoreTests/ServiceTests/ReachabilityServiceTests.swift
JoaoPinhoMinder: Makes sense. Updated conforming with all other suggestions as well ✅

Alfie/AlfieKit/Sources/Core/Services/Reachability/NWPathMonitor+NetworkPathMonitorProtocol.swift
```diff
@@ -0,0 +1,25 @@
+import Combine
+import Models
+import Network
+
+extension NWPathMonitor: NetworkPathMonitorProtocol {
+    public var networkAvailability: AnyPublisher<Bool, Never> {
```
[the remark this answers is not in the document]
JoaoPinhoMinder: I think i got you know. Nice raised concerns 🙏

p4checo (approved):
💪🏼

reisdev (approved):

PR #30, opened 2025-02-12
ALFMOB-89: Error screen header not fixed to the top

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-89

### Description

The divider in the toolbar was not fixed to the top. To resolve this, we now use `frame` and `alignment` to anchor it at the top. However, this also causes views that should be centered within the remaining space to align at the top. To address this, some error views required a `Spacer` at the top and bottom. As part of this fix, a generic `ErrorView was created for use in all error scenarios.

### DIFF
--- a/Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
+++ b/Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
@@ -1,7 +1,0 @@
-import Foundation
-
-struct ErrorButtonConfiguration: Identifiable {
-    let id: String = UUID().uuidString
-    let text: String
-    let action: () -> Void
-}
--- a/Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
+++ b/Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+struct ErrorButtonConfiguration: Identifiable {
+    let id: String = UUID().uuidString
+    let text: String
+    let action: () -> Void
+}
--- a/Alfie/Alfie/Views/ErrorView/ErrorView.swift
+++ b/Alfie/Alfie/Views/ErrorView/ErrorView.swift
@@ -20,7 +20,7 @@ struct ErrorView: View {
     private let titleColor: Color
     private let message: AttributedString?
     private let messageColor: Color
-    private let buttons: [ThemedButton]
+    private let buttonsConfiguration: [ErrorButtonConfiguration]
 
     // MARK: Lifecycle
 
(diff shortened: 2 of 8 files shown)


p4checo (changes requested):
nice improvement!

Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+struct ErrorButtonConfiguration: Identifiable {
+    let id: String = UUID().uuidString
+    let text: String
```
p4checo, line 5: couldn't the `text` be used as `ID`? It's unlikely that we will have multiple buttons with the same text, right?

Also, perhaps call this property `cta` since its effectively the CTA of the button?

Having these automatically generated IDs can have unwanted side effects e.g. if we recreate the array from the exact same values (text), their IDs will always be different, so from SwiftUI's perspective they will be different views, when they could simply be reused.

Alfie/Alfie/Views/ErrorView/ErrorButtonConfiguration.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+struct ErrorButtonConfiguration: Identifiable {
```
p4checo, line 3: since this is only used for `ErrorView`, perhaps we could nest this type inside of it, maybe renaming it to just `ButtonConfiguration`

Alfie/Alfie/Views/ErrorView/ErrorView.swift
```diff
@@ -20,7 +20,7 @@ struct ErrorView: View {
     private let titleColor: Color
     private let message: AttributedString?
     private let messageColor: Color
-    private let buttons: [ThemedButton]
+    private let buttonsConfiguration: [ErrorButtonConfiguration]
```
p4checo, line 23: I think we can leave this as `buttons`. Also, for the current name to be grammatically correct it should be `buttonConfigurations` 🚓

Alfie/Alfie/Views/ErrorView/ErrorView.swift
```diff
@@ -94,8 +94,12 @@ struct ErrorView: View {
             }
 
             VStack(spacing: Spacing.space100) {
-                ForEach(buttons.indices, id: \.self) { index in
-                    buttons[index]
+                ForEach(buttonsConfiguration) { configuration in
+                    ThemedButton(
```
p4checo, line 98: this seems to fit within 120 chars, and we can also use `$0.foo` to make it even more compact if necessary 💅🏼

p4checo (approved):
Nice! Just some minor suggestions

reisdev (approved):

PR #32, opened 2025-02-17
ALFMOB-60: Decouple Navigation from Presentation - Screen type

JoaoPinhoMinder:


JoaoPinhoMinder:
Closing this pull request since it will be deprecated when https://github.com/Mindera/Alfie-iOS/pull/48 is merged. Since this refactor aims to have the application in modules and in order to achieve that, all navigation of the application was revised, meaning the problem stated in this task no longer applies.

PR #34, opened 2025-02-20
ALFMOB-137: Fix CI/CD Release lane

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-137

### Description

After merging the CI/CD implementation, it was observed that the release lane was not progressing due to three distinct issues.

**Bug 1: CI/CD stuck on `Signing GoogleAppMeasurementIdentitySupport.framework`**
The process was waiting for user input, causing it to hang.

**Fix:** Called `setup_ci` before match to create a temporary keychain for CI/CD, ensuring proper signing.

**Bug 2: Build phase script failure `SwiftGen - Copy generated files from OUTPUT_DIR to Project`**
The script was failing because CI/CD modifies the SPM dependencies path.

**Fix:** Updated the script to check if `SPM_CLONED_DEPENDENCIES_PATH` is set and also if is on `Release` configuration. If so, it now uses that value to determine `SWIFTGEN_OUTPUT_DIR` instead of relying on `DERIVED_DATA_PATH`.

**Bug 3: CI/CD not fetching GitHub tags**
Tags were not being fetched, potentially causing issues with versioning.

**Fix:** Explicitly called `git fetch --tags` to ensure tags are retrieved correctly.

### Evidences

Created a commit just for testing purposes where i force release lane to also run, but without tagging and uploading to Testflight and here is the result:
https://github.com/Mindera/Alfie-iOS/actions/runs/13435078399


Alfie/Alfie.xcodeproj/project.pbxproj
```diff
@@ -333,7 +333,7 @@
...
...e DerivedData path\n    DERIVED_DATA_PATH=\"${BUILD_DIR%Build/*}\"\n    SWIFTGEN_OUTPUT_DIR=\"${DERIVED_DATA_PATH}/SourcePackages/plugins/Alfie.output/Alfie/SwiftGenPlugin\"\nfi\n\necho \"Derived Data Path: ${DERIVED_DATA_PATH}\"\necho \"SwiftGen Output Dir: ${SWIFTGEN_OUTPUT_DIR}\"\n\n# Define the original file name\nORIGINAL_FILE_NAME=\"Strings+Generated.swift\"\n\n# Define the new file name\nNEW_FILE_NAME=\"L10n+Generated.swift\"\n\n# Check if the source file exists\nif [ -f \"${SWIFTGEN_OUTPUT_DIR}/${ORIGINAL_FILE_NAME}\" ]; then\n    echo \"Copying ${ORIGINAL_FILE_NAME} to ${DESTINATION_DIR}/${NEW_FILE_NAME}\"\n\n    # Copy the file to the destination with the new name\n    cp \"${SWIFTGEN_OUTPUT_DIR}/${ORIGINAL_FILE_NAME}\" \"${DESTINATION_DIR}/${NEW_FILE_NAME}\"\n    \n    # Uncomment the new file by removing /* and */ markers\n    echo \"Uncommenting the new file: ${DESTINATION_DIR}/${NEW_FILE_NAME}\"\n    sed -i '' 's/\\/\\*//g' \"${DESTINATION_DIR}/${NEW_FILE_NAME}\"  # Remove /* markers\n    sed -i '' 's/\\*\\///g' \"${DESTINATION_DIR}/${NEW_FILE_NAME}\"  # Remove */ markers\nelse\n    echo \"Error: Generated file not found in ${SWIFTGEN_OUTPUT_DIR}\"\n    exit 1\nfi\n";
```
p4checo, line 336: There's a `CI` env var being set up in most CI/CD platforms, and in GH Actions is [no exception](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables). We can use which makes it a bit more clear when looking at the script that is is CI-specific logic.

p4checo (approved):
nice one! just a small suggestion 🤓

Alfie/Alfie.xcodeproj/project.pbxproj
JoaoPinhoMinder: I can use that, but the Release configuration i can't drop, since when running the tests we will have the environment variables as well. Is that ok?

JoaoPinhoMinder: Just remembered one thing, if i use `SPM_CLONED_DEPENDENCIES_PATH` to check if it is set, we guarantee that we have a value to construct the `SWIFTGEN_OUTPUT_DIR`, but if we use a `CI` env var, we will assume that `SPM_CLONED_DEPENDENCIES_PATH` has some value, is that preferred?

p4checo: I would say we check if both `CI` and `SPM_CLONED_DEPENDENCIES_PATH` are set, then we use them. Makes sense?

JoaoPinhoMinder: Updated and here is the new job that also tests the release:
https://github.com/Mindera/Alfie-iOS/actions/runs/13459206702

Had to rebase with main since it was there a PR that added Mocks to a Preview file but without checking if it was in debug mode.

DanielPrimoMinder (approved):

PR #41, opened 2025-03-06
ALFMOB-149: Bag & Wishlist - Redirect to PDP

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-149

### Description

In the **Bag** and **Wishlist** screens, tapping on a product cell now redirects the user to the **PDP** of the selected item, ensuring a smooth transition while preserving selected attributes like size and color.

**Changes & Improvements:**
- Made product cells tappable, including a tap effect for better user feedback.
- Renamed `SelectionProduct` to `SelectedProduct` for better clarity.
- Updated `SelectedProduct` to include `Product` and `Product.Variant` properties.
- Implemented a mechanism to open **PDP** with a `SelectedProduct`.
- Fixed an issue in `buildColorAndSizingSelectionConfigurations` where the `Product.Variant` parameter was not being properly used.

### Evidences

https://github.com/user-attachments/assets/5ecbc6b3-2f17-43d3-a804-3b3dae9d4d51


Alfie/Alfie/Navigation/ViewFactory.swift
```diff
@@ -157,6 +157,23 @@ final class ViewFactory: ViewFactoryProtocol {
                         )
                     )
                 )
+
+            case .selectedProduct(let selectedProduct):
+                ProductDetailsView(
+                    viewModel: ProductDetailsViewModel(
+                        productId: selectedProduct.product.id,
```
p4checo, line 164: why not inject just a `SelectedProduct`? 🤔  (instead of the first 3 parameters that are derived from it)

Alfie/Alfie/Views/ProductDetails/ProductDetailsViewModel.swift
```diff
@@ -87,16 +88,32 @@ final class ProductDetailsViewModel: ProductDetailsViewModelProtocol {
...
+        dependencies: ProductDetailsDependencyContainer
+    ) {
         self.productId = productId
+        self.initialSelectedProduct = selectedProduct
         baseProduct = product
         self.dependencies = dependencies
 
-        if let baseProduct {
+        switch (product, selectedProduct) {
+        case (.some(let product), .none):
```
p4checo, line 103: please don't use explicit `Optional` cases in pattern matching (i.e. `.some` / `.none`), use `(let product, nil)` instead, in this case

Alfie/Alfie/Views/ProductDetails/ProductDetailsViewModel.swift
```diff
@@ -87,16 +88,32 @@ final class ProductDetailsViewModel: ProductDetailsViewModelProtocol {
         product?.priceType
     }
 
-    init(productId: String, product: Product?, dependencies: ProductDetailsDependencyContainer) {
+    init(
+        productId: String,
+        product: Product?,
```
p4checo, line 93: why can we have both a `Product` and a `SelectedProduct` being passed in at the same time, and both optional? What scenarios are we covering?

Alfie/AlfieKit/Sources/Models/Models/Base/SelectedProduct.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+
+public struct SelectedProduct {
+    public let product: Product
+    public let selectedVariant: Product.Variant
+
+    public var name: String {
```
p4checo, line 7: Could we use`@dynamicMemberLookup` (and respective subscript(s)) instead?

This model is a bit messy, because some things hook into the `Product`, others into the `selectedVariant`, but then the `Product` also has variants... 😕

Alfie/AlfieKit/Sources/Models/Models/Base/SelectedProduct.swift
```diff
@@ -0,0 +1,53 @@
...
+        self.selectedVariant = selectedVariant ?? product.defaultVariant
+    }
+}
+
+extension SelectedProduct: Identifiable, Equatable, Hashable {
+    public var id: String {
+        "\(product.id)-\(selectedVariant.sku)"
+    }
+
+    public static func == (lhs: SelectedProduct, rhs: SelectedProduct) -> Bool {
```
p4checo, line 46: this is usually not a good practice, because it breaks the classic assumption that equality means the whole value is equal, not just its ID (same thing for `Hashable` below)

p4checo (changes requested):

Alfie/Alfie/Views/ProductDetails/ProductDetailsViewModel.swift
JoaoPinhoMinder: when we have only `productId`, we need to fully fetch the product remotely. If we have just a `Product`, means that we will have a default variant being selected automatically, but if we have a `SelectedProduct`, means that we can show a product variant pre-selected or simply a product without any pre-selection, this is not yet seen in this PR, since that's part of https://github.com/Mindera/Alfie-iOS/pull/42. I already improved to have an enum instead of allowing all of them being passed as optional, double check to see if it makes sense for you.

Alfie/AlfieKit/Sources/Models/Models/Base/Media.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-public enum Media {
+public enum Media: Equatable, Hashable {
```
p4checo, line 3: `Hashable: Equatable`, so conforming to `Equatable` is redundant here (multiple cases)

Alfie/AlfieKit/Sources/Models/Models/Base/MediaImage.swift
```diff
@@ -13,4 +13,16 @@ public struct MediaImage {
         self.mediaContentType = mediaContentType
         self.url = url
     }
+
+    public static func == (lhs: MediaImage, rhs: MediaImage) -> Bool {
```
p4checo, line 17: I think we can let the compiler generate these conformances, since all properties are now considered? (multiple cases)

Alfie/Alfie/Views/ProductDetails/ProductDetailsViewModel.swift
```diff
@@ -89,31 +89,36 @@ final class ProductDetailsViewModel: ProductDetailsViewModelProtocol {
     }
 
     init(
-        productId: String,
-        product: Product?,
-        selectedProduct: SelectedProduct? = nil,
+        productKind: ThemedProductDetailsScreen,
```
p4checo, line 92: the type of this parameter seems a bit off, can we perhaps rename it? I don't know, `ProductDetailsConfiguration`?

Alfie/AlfieTests/Features/ProductDetailsViewModelTests.swift
```diff
@@ -873,8 +873,13 @@ final class ProductDetailsViewModelTests: XCTestCase {
     // MARK: - Helper methods
 
     private func initViewModel(productId: String = "", product: Product? = nil) {
```
p4checo, line 875: why not change this (helper) API to receive a product Type/Kind/Configuration instead?

p4checo (changes requested):

Alfie/AlfieKit/Sources/Models/Models/Base/SelectedProduct.swift
p4checo: The idea was to simplify, not add complexity. If you think it adds complexity, then we can leave it as is.

Alfie/Alfie/Views/ProductDetails/ProductDetailsViewModel.swift
p4checo: the parameter name can probably be just `configuration` as the type name itself provides extra context 💅🏼

p4checo (approved):

reisdev (approved):

PR #42, opened 2025-03-11
ALFMOB-162: Improve Wishlist Functionality on PLP & Wishlist

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-162

**Depends on:** https://github.com/Mindera/Alfie-iOS/pull/41

### Description

- Created `WishlistProduct` and `BagProduct` inheriting from `SelectedProduct`.
- Made `WishlistProduct` unique by **product ID and color ID**.
- Made `BagProduct` unique by **product ID and variant SKU**.
- Removed size information from `WishlistView` and `ProductListingView`.
- Updated **"Add to Bag"** action in `WishlistView` to redirect users to **PDP** for size selection instead of adding directly to the bag.
- Disabled **"Add to Bag"** button on **PDP** until both **color** (pre-selected) and **size** have been selected.

### Evidences

https://github.com/user-attachments/assets/b2555819-13f8-49ed-a844-ca9ebcf487d9

PR #44, opened 2025-03-13
ALFMOB-158: Update PDP Button Copy Based on Item State

JoaoPinhoMinder:
### Ticket

https://mindera.atlassian.net/browse/ALFMOB-158

**Depends on:** https://github.com/Mindera/Alfie-iOS/pull/42

### Description

- Added `productsPublisher` to both `WishlistService` and `BagService` to enable real-time updates of button titles based on item state.
- Updated all relevant view models to observe `productsPublisher` instead of using the previous `getWishlistContent` and `getBagContent` methods.
- Adjusted button titles dynamically to reflect the correct state (`"Add to Bag"` / `"Remove from Bag"` and `"Add to Wishlist"` / `"Remove from Wishlist"`).
- Ensured that changing color resets both buttons, while changing size only resets the `"Add to Bag"` button (since wishlist state is unaffected by size).
- Added necessary localized strings to `L10n` table.

### Evidences

https://github.com/user-attachments/assets/eedeafc1-31bb-499e-b59b-ede39c0b886b


p4checo:
This PR seems to build on top of #42, so ideally we should have that not on the PR's description, and point it to the #42 PR which then creates a merge train 🚋

amccall-mindera:
## Board Review — ALFMOB-158: PDP Button Copy

### 🔴 Blocked — Depends on ALFMOB-162 rewrite

This draft PR depends on PR #42 (ALFMOB-162) which requires a full rewrite due to merge conflicts with the modular architecture refactor (ALFMOB-184).

**Next steps:** Once ALFMOB-162 is reimplemented on the new architecture, this feature should also be reimplemented fresh on top of it. The feature requirements (dynamic Add/Remove button text based on item state) remain valid — use this PR as a spec reference.

cc @JoaoPinhoMinder

PR #54, opened 2025-12-18
ALFMOB-202: Fix iOS 26 liquid glass effect bug

paulcristian-mindera:
| With Glass on Tab Bar | Without Glass| 
| --------------- | ------------ |
| simulator_screenshot_500BEDC4-B044-49D9-9430-D96DEBA2D189 | simulator_screenshot_BFDE279C-F5FC-4E29-BDDB-F0977AB92318|


amccall-mindera (approved):

amccall-mindera (commented):
tested and good to go

Alfie/Alfie.xcodeproj/project.pbxproj
```diff
@@ -62,7 +62,6 @@
 				Alfie.xctestplan,
 				"Configuration/Debug/GoogleService-Info.plist",
 				"Configuration/Debug/GoogleService-Info.plist.secret",
-				"Configuration/Release/GoogleService-Info.plist",
```
amccall-mindera: Was this deliberate?

amccall-mindera:
## Code Review — ALFMOB-202: iOS 26 TabBar Fix

### 🟡 High — `.toolbar(.hidden, for: .tabBar)` modifier placement

The `.toolbar(.hidden, for: .tabBar)` modifier appears to be applied to the `ForEach` inside the `TabView` rather than the `TabView` itself in `RootTabView.swift`.

**Current:**
```swift
TabView(selection: $viewModel.selectedTab) {
    ForEach(viewModel.tabs, id: \.self) { tab in
        // ...
    }
    .toolbar(.hidden, for: .tabBar)  // Applied to ForEach
}
```

**Suggested:**
```swift
TabView(selection: $viewModel.selectedTab) {
    ForEach(viewModel.tabs, id: \.self) { tab in
        // ...
    }
}
.toolbar(.hidden, for: .tabBar)  // Applied to TabView
```

This may not reliably hide the system tab bar across all iOS versions (16.0+), potentially causing the liquid glass effect to still appear on some devices. Could you verify this works correctly on both iOS 16 and iOS 26 simulators?

Overall approach is sound — hiding the system tab bar and using a custom one is the right call for this.

PR #60, opened 2025-12-23
ALFMOB-204: Added UI Test example

paulcristian-mindera:
https://github.com/user-attachments/assets/7ef91b72-c50d-4912-91fc-f2c5c3565330


amccall-mindera (approved):

amccall-mindera:
## Code Review — ALFMOB-204: UITesting Example

This is meant to serve as a **starting point template** for the team, so code quality here sets the standard. Three issues found:

### 🔴 Critical — `app.activate()` should be `app.launch()`

In `AlfieUITests.swift:20`, `app.activate()` only brings an already-running app to the foreground — it does **not** launch the app if it isn't running. This means the test will fail on a clean test run.

**Fix:** Change `app.activate()` to `app.launch()` to ensure the app starts in a fresh, known state for each test.

### 🟡 High — Hard-coded brand name makes test brittle

Line 23 taps on `"04AugPhilos"` which is specific test/sample data. If this brand is removed or renamed, the test breaks silently. For a template, this should demonstrate a more robust approach.

**Suggestion:** Either use a more generic selector (e.g., find the first available brand), or add a comment/doc explaining the data dependency.

### 🟠 Medium — Xcode UI Recorder tokens left in

Lines 21-23 contain `/*@START_MENU_TOKEN@*/.../*@END_MENU_TOKEN@*/` auto-generated placeholders. These should be cleaned up for readability, especially in a template that others will copy.

**Fix:** Replace with explicit, readable selectors, e.g.:
```swift
app.otherElements["shop-tab"].images["storefront"].tap()
app.scrollViews.buttons["Brands"].tap()
```

PR #70, opened 2026-05-15
ALFMOB-150: Persist Wishlist & Bag

khoinguyen-mindera:
## Ticket
[ALFMOB-150](https://mindera.atlassian.net/browse/ALFMOB-150)

## Summary
- Persist Wishlist and Bag across app sessions using `UserDefaults`.
- New `WishlistStoreProtocol` / `BagStoreProtocol` (domain layer) decouple services from the storage backend, making future backend sync a drop-in replacement.
- Shared `UserDefaultsStore` (infrastructure) implements both protocols and maps domain `SelectedProduct` ↔ internal `PersistedProductDTO` (Codable). Keeps `Product`/`PriceType` domain models free of `Codable` conformance.
- `WishlistService` / `BagService` now load on init and save on every add/remove; public protocol surfaces unchanged so no consumer (ViewModels, PDP, PLP) had to be touched.
- Single shared `StorageKey` enum exposing `.wishlistItems` and `.bagItems` — no user-scoped namespacing, so login/logout and guest usage all share the same persisted data.
- Data is plain JSON, unencrypted, no item limit, removed on app uninstall.
- Added unit tests for `WishlistService`, `BagService`, `UserDefaultsStore`, and the DTO round-trip, including memory-leak tracking via `trackForMemoryLeak`.

## Screenshot

[ALFMOB-150]: https://mindera.atlassian.net/browse/ALFMOB-150?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amccall-mindera:
Architecture and test coverage both look solid — the domain-protocol + DTO split is the right call and the tests are thorough. A few things worth resolving before merge, mostly around the bag side of the shared DTO and a couple of silent-failure cases.

Main thing: the simplification of dropping full variant fidelity makes sense for the wishlist (row is display-only, PDP re-fetches), but the same DTO is doing duty for the bag, and the bag will feed checkout. The DTO drops `currencyCode` and the numeric `amount` — keeping only `amountFormatted`. After relaunch, you can render the bag row but can't compute a subtotal, apply a discount, or convert currency without a refetch. Is the plan that the bag also re-hydrates on open (same as PDP), or is there a follow-up to extend the DTO for bag-side fidelity?

Smaller stuff inline.

khoinguyen-mindera:
@amccall-mindera Thank you 

I have replaced `PersistedPriceTypeDTO` (formatted strings only) with `PersistedPriceDTO` / `PersistedMoneyDTO` that mirror `Price` / `Money` field-for-field. Numeric `amount` and `currencyCode` are now persisted alongside `amountFormatted`, so checkout-side math will read real values after a cold launch — no more stubbed zeros.

Alfie/AlfieKit/Sources/Core/Services/Wishlist/WishlistService.swift
```diff
@@ -1,23 +1,27 @@
...
+    private let store: WishlistStoreProtocol
 
-    public init() { }
+    public init(store: WishlistStoreProtocol) {
+        self.store = store
+    }
 
     public func addProduct(_ product: SelectedProduct) {
+        var products = store.load()
         guard !products.contains(where: { $0.id == product.id }) else { return }
```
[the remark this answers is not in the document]
khoinguyen-mindera: Good catch — real inconsistency. Pre-existing in main: `addProduct` dedupes by variant (`SelectedProduct.id`) while `removeProduct(withId:)` removes by product (`product.id`). This PR routes both through the new store but doesn't change either predicate.

The intended UX is wishlist-as-products (user adds a product, picks size/colour later when they navigate to PDP), so the fix is to also dedupe `addProduct` by `product.id`. Out of scope here — will raise a follow-up ticket.

Alfie/AlfieKit/Sources/Core/Services/Persistence/UserDefaultsStore.swift
```diff
@@ -0,0 +1,28 @@
...
+        guard
+            let data: Data = userDefaults.value(for: storageKey),
+            let dtos = try? JSONDecoder().decode([PersistedProductDTO].self, from: data)
+        else {
+            return []
+        }
+        return dtos.map(\.selectedProduct)
+    }
+
+    public func save(_ products: [SelectedProduct]) {
```
amccall-mindera, line 23: Every `addProduct` / `removeProduct` in `BagService` / `WishlistService` does `store.load()` → mutate → `store.save(...)` on the caller's thread. Since those services are typically invoked from `@MainActor` view models, the full JSON decode/encode runs on the main thread on every mutation — frame drops will start showing up as bags grow.

Suggested shape: hold an in-memory `[SelectedProduct]` cache inside each service (hydrated once at init from the store), serve reads from the cache, and only encode+write to `UserDefaults` from a background context. Making the services `actor`s would solve this cleanly — encode+write moves off main automatically, concurrent mutations are serialised (closing a lost-update window), and the async API is what you'd want anyway when the backend-sync swap lands.

amccall-mindera (changes requested):
Clean layering and good test coverage on the persistence boundary. One blocker before merge — see the inline comment on `UserDefaultsStore.save`.

@khoinguyen-mindera what are you thoughts on this? Too complex, the efficiency gains too minimal?

Alfie/AlfieKit/Sources/Core/Services/Persistence/UserDefaultsStore.swift
khoinguyen-mindera: - `BagService`/`WishlistService` are now `actor`s with the in-memory cache added back. Reads and writes are offloaded from the main thread (encode/decode runs on the actor), and concurrent mutations are serialised. UserDefaults stays the durable store (write-through), so the AC is unaffected.

- Bring back in-memory cache in service layer

amccall-mindera (approved):
Thanks Khoi — actor + write-through cache is exactly the right shape here. Main-thread JSON cost is gone, reads are O(1), and concurrent mutations are now serialised for free. Approving.

PR #78, opened 2026-06-09
ALFMOB-333: Implement Search via BFF

khoinguyen-mindera:
## Ticket
[ALFMOB-333](https://mindera.atlassian.net/browse/ALFMOB-333)

## Summary
- Add `SearchProductsQuery` and call the BFF `searchProducts` query for search, reusing `ProductListItemFragment` and a parallel response converter.
- Send the predefined `platform` on `searchProducts` (the BFF requires it, unlike `productList` which defaults to Shopify server-side).
- Search results render via the existing Product List flow in `.searchResults` mode (navigation push) — no inline results UI.
- Remove the obsolete suggestion/autocomplete path (`GetSuggestions`, `SearchSuggestion*`, the inline suggestion UI); `SearchViewState` reduced to `empty`/`recentSearches`.
- Trim whitespace-only submissions so they don't fire a request or pollute recent searches.

### Refactor: split product browse vs search into dedicated services
- Replace the muxed `productListing(categoryId:query:)` (which did two things and allowed invalid arg combos) with single-purpose methods at every layer:
  - `BFFClientService.productList(collectionHandle:)` + `searchProducts(searchTerm:)`
  - `ProductService.productList(collectionHandle:)` (browse)
  - revived `SearchService.searchProducts(searchTerm:)` (search) — also the future home for suggestions ([ALFMOB-404](https://mindera.atlassian.net/browse/ALFMOB-404))
  - `ProductListingService.productListPage` / `searchPage`
- The browse-vs-search decision now lives in `ProductListingViewModel` (`switch mode`), the component that knows the flow.
- Updated mocks + tests; added `SearchServiceTests`.

## Screenshot

https://github.com/user-attachments/assets/bdd32339-4c96-427f-8f90-6ecefe0f3dd3

[ALFMOB-333]: https://mindera.atlassian.net/browse/ALFMOB-333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ALFMOB-404]: https://mindera.atlassian.net/browse/ALFMOB-404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alfie/AlfieKit/Sources/Search/UI/SearchViewModel.swift
```diff
@@ -112,38 +78,16 @@ extension SearchViewModel {
 
 extension SearchViewModel {
     public func onSubmitSearch() {
-        guard isSearchSubmissionAllowed else {
+        let term = searchText.trim()
```
amccall-mindera, line 81: `onSubmitSearch` correctly trims and rejects whitespace-only input, but `isSearchSubmissionAllowed` (line 36) checks `!searchText.isEmpty` without trimming. A whitespace-only string like `"   "` leaves the submit button appearing active, then tapping it silently does nothing.

Fix — trim in `isSearchSubmissionAllowed` to keep button state consistent with submission behaviour:

```swift
public var isSearchSubmissionAllowed: Bool {
    !searchText.trim().isEmpty
}
```

Test: assert `sut.isSearchSubmissionAllowed == false` when `sut.searchText = "   "`.

Alfie/AlfieKit/Sources/Search/UI/SearchViewModel.swift
```diff
@@ -112,38 +78,16 @@ extension SearchViewModel {
...
-
-    public func onTapOpenBrands() {
-        guard let url = ThemedURL.brands.internalUrl else { return }
-        ExternalAppLauncher.open(url: url)
+        dependencies.recentsService?.add(.text(value: term))
+        dependencies.analytics.trackSearch(term: term)
+        navigate(.searchIntent(.productListing(searchTerm: term, category: nil)))
     }
 
     public func viewDidAppear() {
```
amccall-mindera, line 90: AC: [ALFMOB-333](https://mindera.atlassian.net/browse/ALFMOB-333) — "Recent searches and the top search bar continue to behave as today."

`viewDidAppear` is the trigger that surfaces recent searches every time the screen opens, but the Lifecycle section in `SearchViewModelTests` only covers `viewDidDisappear` and `closeSearch`. A regression in this function would go undetected.

Suggest adding:

```swift
func test_viewDidAppear_withRecentSearches_setsRecentSearchesState() {
    mockRecentsService.recentSearches = [.text(value: "polo")]
    let sut = makeSUT()
    sut.state = .empty
    sut.viewDidAppear()
    XCTAssertEqual(sut.state, .recentSearches)
}

func test_viewDidAppear_withoutRecentSearches_doesNotChangeState() {
    let sut = makeSUT()
    sut.viewDidAppear()
    XCTAssertEqual(sut.state, .empty)
}
```

Alfie/AlfieKit/Sources/Search/UI/SearchViewModel.swift
khoinguyen-mindera: Updated

Alfie/AlfieKit/Sources/Search/UI/SearchViewModel.swift
khoinguyen-mindera: Added test

amccall-mindera (approved):
Clean implementation — all AC satisfied, search correctly routes through the BFF via `searchProducts`, PLP handles both listing and search modes, submission validation trims whitespace, and recent searches remain local-only. Previous review comments addressed. No new findings.

PR #79, opened 2026-06-10
ALFMOB-385: Locale-aware price formatting for BFF Money

khoinguyen-mindera:
## Ticket
[ALFMOB-385](https://mindera.atlassian.net/browse/ALFMOB-385)

## Summary
- Add `CurrencyFormatter` (Model) — locale/currency-symbol aware price strings via `NumberFormatter`/`.currency`, plus per-currency minor-unit scaling using the currency's ISO-4217 exponent.
- Route both BFF converters' `toDomainMoney()` through it: parse the BFF `Double` once into a clean `Decimal`, then derive `amountFormatted` (e.g. `£10.00` instead of `10.00 GBP`) and the integer minor-unit `amount`.
- Fix the hardcoded `×100` scaling so 0-decimal (JPY) and 3-decimal (KWD/BHD) currencies are correct; guard non-finite amounts.
- Add `ModelTests` target with per-currency formatting/scaling/rounding tests (pinned locales); add converter-boundary tests and make assertions locale-independent.

## Screenshot

Simulator Screenshot - iPhone 17 Pro - 2026-06-11 at 10 37 57

[ALFMOB-385]: https://mindera.atlassian.net/browse/ALFMOB-385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alfie/AlfieKit/Sources/Model/Formatting/CurrencyFormatter.swift
```diff
@@ -0,0 +1,42 @@
...
+        currencyCode: String,
+        locale: Locale = .current
+    ) -> String {
+        amount.formatted(.currency(code: currencyCode).locale(locale))
+    }
+
+    /// Minor-unit exponent for a currency (GBP=2, JPY=0, KWD=3), as reported by `NumberFormatter`
+    /// for the given code. Unrecognised codes fall back to 2 (NumberFormatter's default).
+    public static func minorUnitDigits(for currencyCode: String) -> Int {
+        let formatter = NumberFormatter()
```
amccall-mindera: `NumberFormatter` is expensive to initialise (loads ICU/locale data from the OS) and is being allocated on every call to `minorUnits()`. A PLP page with 20 products × 2 prices (priceRange low/high) triggers 40+ allocations per page load, all for the same currency code.

Cache by currency code:

```swift
private static var minorUnitDigitsCache: [String: Int] = [:]

public static func minorUnitDigits(for currencyCode: String) -> Int {
    if let cached = minorUnitDigitsCache[currencyCode] { return cached }
    let formatter = NumberFormatter()
    formatter.numberStyle = .currency
    formatter.currencyCode = currencyCode
    let digits = formatter.maximumFractionDigits
    minorUnitDigitsCache[currencyCode] = digits
    return digits
}
```

Note: if BFF responses are ever parsed concurrently, guard the cache with `nonisolated(unsafe)` + a lock, or replace with a pre-populated `static let` dictionary of known ISO-4217 codes.

**Test:** In `CurrencyFormatterTests`, call `CurrencyFormatter.minorUnitDigits(for: "GBP")` twice and assert both return `2` — the existing `test_minor_unit_digits_per_currency` already covers correctness; no extra assertion needed beyond proving the second call doesn't crash or regress.

khoinguyen-mindera: Done in 09bdfe9 — memoised by currency code. Used `OSAllocatedUnfairLock` (iOS 16+) rather than a bare dict, since converters run off-main on the cooperative pool and concurrent fetches can hit this in parallel. Added a 1,000-iteration concurrent-access test to guard against the lock being dropped.

amccall-mindera (approved):
Approving — all ALFMOB-385 acceptance criteria verified. The new CurrencyFormatter is locale-aware (.currency + currencyCode) and derives minor-unit scale per currency (GBP=2, JPY=0, KWD=3); toDomainMoney() formats at the BFF boundary via a float→String→Decimal hop that avoids binary-float noise, with non-finite input guarded. Test coverage is thorough: per-currency symbol/grouping/decimals, exact minor-unit scaling for non-2dp currencies, negatives, half-away rounding, and >Int32 overflow. All three prior review threads are addressed and verified at head — the NumberFormatter caching is now memoised behind OSAllocatedUnfairLock (concurrency-safe for off-main converters), and the weak/ silent-zero test patterns are replaced with exact assertions and fail-loud XCTUnwrap. The stored amountFormatted remains a ticket-sanctioned smaller-change tradeoff (Problem #3 deferred).

PR #99, opened 2026-07-23
ALFMOB-387: Load Shop categories from BFF mainMenu, replace static mock

khoinguyen-mindera:
## ALFMOB-387 — Shop categories via BFF `mainMenu`

[ALFMOB-387](https://mindera.atlassian.net/browse/ALFMOB-387)

Replaces the temporary in-memory category tree in `getHeaderNav` with a real BFF call to the new `mainMenu(handle:platform:)` query, mapping the Shopify `Menu`/`MenuItem` tree → domain `[NavigationItem]` at the boundary (mirrors the `ProductListing`/`ProductDetails` converters).

### What changed
- **New query** `MainMenuQuery` (3-level selection — matches Shopify's menu depth cap) + schema synced from BFF `main` via `sync-bff-schema.sh` (regenerated Apollo types/mocks).
- **New converter** `MainMenu+Converter.swift`: `Menu` → `[NavigationItem]`. Each leaf resolves to `/<collection-handle>` (last path segment, minus `?`query/`#`fragment, lowercased) so the existing `didSelectCategory` → PLP flow works unchanged. Leaves with no handle and no children are dropped.
- **`BFFClientService.getHeaderNav`** rewired to `mainMenu` + converter; **temporary static mock removed**. `NavigationHandle.header → "main-menu"`.
- **Navigation flow:** the Categories row chevron now shows **only when an item has sub-items**; leaves render chevron-less and go straight to the PLP.
- **Tests:** 11 converter unit tests (`MainMenuConverterTests`).

### Acceptance criteria
- ✅ Shop → Categories loads the tree from the BFF (no static mock).
- ✅ Selecting a leaf opens a PLP for that collection handle (unit-verified; see note).
- ✅ No references to the temporary in-memory mock remain.

### Verification
`./Alfie/scripts/verify.sh --skip-integration` → ✅ build + unit tests pass.

### Notes for reviewers
- **Integration not run here** — the real Shopify menu URL → collection-handle shape is only exercised end-to-end by the integration suite against a live BFF. Worth a manual Categories→PLP smoke test.
- **Schema diff is large** — `sync-bff-schema.sh` refreshes the whole committed `schema.graphqls` to mirror BFF `main`; Apollo only generated types reachable from operations (`MainMenuQuery` + `Menu`/`MenuItem`), no Cart types.
- **Follow-up:** if `/brands` or `/store-services` (SpecialCategories) appear in `main-menu`, they arrive as leaves and now render chevron-less despite navigating to a screen — confirm with design (they may be separate tabs). Deep-link slug follow-up tracked in ALFMOB-386.

[ALFMOB-387]: https://mindera.atlassian.net/browse/ALFMOB-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

## Also in this PR — Debug: switch BFF commerce platform (DEBUG only)

Adds a **Commerce Platform** row to the Debug Menu to flip the BFF `platform` argument (`shopify` / `bigcommerce`) at runtime.

- `BFFPlatform.predefined` is now DEBUG-gated: release builds always send `.shopify`; DEBUG builds honor `BFFPlatformDebugStore` (UserDefaults). Applies to the **next request — no restart** (unlike the Environment switch).
- New `PlatformSelectionView` + view-model (radio list), mirrors the Feature Toggle screen; the menu link is `#if DEBUG`-gated.
- Tests: `BFFPlatformTests` (override read/persist/default) + `PlatformSelectionViewModelTests`.

_Note: bundled here at request; it's an independent concern from the categories work._

---

## Also in this PR

**Send `platform` on `productList`** (`0e3b232`) — `ProductListQuery` gained a `$platform` arg; `productList` now sends `BFFPlatform.predefined` (like `productDetails`/`searchProducts`) instead of relying on the BFF's Shopify default. This makes the debug platform override actually apply to category/collection listings. Behavior change: all product-list calls now always send a platform.

**Pull-to-refresh on Shop categories** (`c52f66c`) — `.refreshable` on the categories list re-fetches `mainMenu` via a new `CategoriesViewModel.refresh()`. Keeps the current list visible during refresh (own spinner, no placeholder flip) and preserves it if the re-fetch fails. Applies at every drill-down level; no-op on static sub-category screens. +3 unit tests.


Prathimaboggu:
Nice work — the architecture is right and the `CancellableBox` exactly-once fix is solid. A few things I'd like to resolve before merge:

### 1. Non-collection menu links become broken PLPs

`collectionHandleURL` takes the last path segment of *any* URL, and `makeNavigationItem` stamps `type: .listing` unconditionally. On a stock Shopify `main-menu` that gives:

| Menu URL | Result |
|---|---|
| `/collections/womens-tops` | `/womens-tops` ✅ |
| `/pages/store-locator` | `/store-locator` → empty PLP ❌ |
| `/blogs/news` | `/news` → empty PLP ❌ |
| `/products/some-shirt` | `/some-shirt` → empty PLP ❌ |

Two knock-ons:
- The log in `BFFClientService` (*"likely non-collection links (e.g. `/pages/*`, absolute urls)"*) says these are dropped — they aren't. Only path-less URLs are.
- `didSelectCategory`'s `switch category.type` webview branch is now unreachable for BFF items, so the existing `.page` / `.externalHttp` fallback is silently bypassed.

Suggest switching on the Shopify route prefix rather than taking the last segment blindly:

```swift
let segments = URLComponents(string: trimmed)?.path.split(separator: "/").map(String.init) ?? []
switch segments.first {
case "collections":                      // → .listing, last segment as handle
case "pages", "blogs":                   // → .page, keep full path
case "products":                         // → .product
case .some where segments.count == 1:    // → .listing (bare handle)
default:                                 // drop
}
```

Note `test_all_items_map_to_listing_type` currently encodes this as intended behaviour, so it won't catch it.

### 2. The menu is effectively cached for the app's lifetime

`.default` (`returnCacheDataElseFetch`) + no TTL on the normalized cache + `loadItems`'s `guard !state.isSuccess` means a merchandising change is invisible until the user pulls to refresh or relaunches. Worth `.fetchIgnoringCacheData` on the first load per app session.

### 3. Pull-to-refresh may be unreachable from the error state

`.refreshable` is on `scrollContent`, but `errorView` is an `.overlay` on top of it with no `.allowsHitTesting(false)` — a drag starting on it likely won't reach the ScrollView's pan recogniser. `ErrorView` has no retry button here, so the error state may have no recovery path at all. Worth a manual check.

### Smaller

- `getHeaderNav`'s `includeSubItems` / `includeMedia` are now completely unused — drop them or honour them.
- `didSelectCategory` matches `SpecialCategories` *before* `hasSubCategories`, so a `/brands` item that came back *with* children could never drill down.
- `Menu.handle` / `title` are queried but unused — `title` could feed the screen title.
- `loadItems()` and `refresh()` can race; guards prevent success→error, but not stale-success overwriting fresh-success.
- `CancellableBox` never clears `cancellable` / `resumeOnCancel` after a normal resume, so it retains a continuation-capturing closure.
- `BFFPlatformDebugStore` is public and compiles in release, where `predefined` ignores it — consider `#if DEBUG` on the store too.
- Debug strings are hardcoded where the sibling Feature Toggle screen uses `L10n`.
- `BFFPlatformTests` / `PlatformSelectionViewModelTests` mutate `UserDefaults.standard` — save/restore is done correctly, but an injected suite would remove the hazard.

### Schema / scope

Checked the `OmniProduct.rating` + `Rating` removal — no Swift references anywhere, safe. Cart additions are inert. `productList` now sending `platform` as a required `String!` is a real BFF contract change, correctly called out.

Test coverage is strong (18 converter cases, the Copilot race regression, the refresh-keeps-list cases). Given finding #1, I'd do the manual Categories → PLP smoke test **before** merge rather than after — the integration suite is the only thing that exercises the real menu URL shape.

Prathimaboggu:
Thanks — re-reviewed at `a8702b9`. The three blocking findings are properly addressed:

- **Route-prefix mapping** is in and matches what I'd hoped for. `test_all_items_map_to_listing_type` (which encoded the old behaviour) is gone, replaced by per-route assertions plus a drop-the-unknown case. The misleading `BFFClientService` log is accurate now.
- **Caching** — dropping `forceRefresh` entirely and always sending `.fetchIgnoringCacheData` is cleaner than the per-session flag I suggested. `loadItems`' `!state.isSuccess` guard now just means "don't refetch on tab re-entry", which is fine.
- **Error recovery** — Retry button resolves it.

Two things I'd still like before merge, then some carry-over nits.

### 1. The host is still discarded, so external links hit the same failure mode as `/pages/*` did

`URLComponents(string:).path` drops the host, and nothing in the converter ever produces `.externalHttp` (it only appears in `didSelectCategory`'s consuming switch). So an off-site menu link still resolves to an on-site destination:

| Menu URL | Result |
|---|---|
| `https://www.instagram.com/alfie` | path `/alfie` → single segment → `.listing` `/alfie` → empty PLP ❌ |
| `https://partner.example.com/pages/lookbook` | `.page` `/pages/lookbook` → webview rebuilt on `ThemedURL.hostWithPortComponent`, i.e. **our** host ❌ |

The second is the worse one: `didSelectCategory` reconstructs the URL from path components against our own host, so a foreign link silently loads the wrong page rather than failing visibly.

`test_absolute_url_reduces_to_last_path_segment` uses `shop.example.com`, so nothing currently pins the foreign-host case either way.

Suggest resolving the host in `menuDestination` and letting genuinely external links through as `.externalHttp` with the absolute URL intact:

```swift
guard let components = URLComponents(string: trimmed) else { return nil }
// nil host = relative url = our own shop
if let host = components.host, host != <shop host> {
    return (.externalHttp, trimmed)
}
```

…and `didSelectCategory`'s webview branch should use `category.url` verbatim when the type is `.externalHttp`, instead of re-hosting the path. A test with a foreign host would lock it in.

### 2. `/collections/<handle>/<tag>` picks the tag as the handle

`case "collections"` takes `segments.last`, but Shopify's tag-filtered collection links are three segments — `/collections/all/sale` yields handle `sale`, not `all`. `segments[1]` is the handle (dropping the tag is fine for now, or thread it through as a query param). Related: a bare `/collections` currently yields handle `collections`. No test covers the 3-segment shape.

### Carry-over from the last round (not blocking, but still open)

- `getHeaderNav`'s `includeSubItems` / `includeMedia` — `forceRefresh` got dropped, but these two are still passed and still ignored in the body.
- `didSelectCategory` still matches `SpecialCategories` before `hasSubCategories`, so a `/brands` item that came back *with* children can't drill down.
- `Menu.handle` / `title` still queried and unused — `CategoriesViewStateModel.title` is left at its `""` default in `fetchNavigationItems`.
- Stale-success overwrite: `state = .success(...)` is still unconditional; only the error and empty paths got the `!state.isSuccess` guard.
- `CancellableBox.resumeOnce` still doesn't clear `cancellable` / `resumeOnCancel`, so the box holds the continuation closure and the Apollo cancellable after a normal resume.
- `PlatformSelectionView` strings still hardcoded ("Choose Platform" + the subtitle) where the sibling Feature Toggle screen uses `L10n`.
- `BFFPlatformTests` / `PlatformSelectionViewModelTests` still mutate `UserDefaults.standard` — save/restore is correct, just a shared-state hazard.
- `BFFPlatformDebugStore` release-compiled — the new doc comment justifies it; happy to leave as-is.

### Nits on the new code

- Retry doesn't flip state to `.loading`, so tapping it gives no feedback until it resolves.
- The error overlay still has no `.allowsHitTesting(false)`, so pull-to-refresh from the error state is likely still swallowed — moot now that Retry exists, but worth the one-liner.

Still worth the manual Categories → PLP smoke test before merge (the PR body notes integration wasn't run).

khoinguyen-mindera:
Thanks @Prathimaboggu — review addressed. Status:

Substantive items — done:

#1 (non-collection links → empty PLP): resolved by treating the Shop menu as product-collections-only. Page/blog/product and external links are bad data for this surface and are now dropped at the converter, and we've pointed the query at a dedicated curated product-category menu instead of the general main-menu. Verified with a manual Categories → sub-menu → PLP smoke test against the new menu — drill-down and PLP resolve correctly.
#2 (/collections/<handle>/<tag>): fixed — handle is segments[1] (tag dropped); bare /collections dropped. Tests added.
Carry-overs: includeSubItems/includeMedia removed; hasSubCategories now matched before special-casing; stale-success covered by the single-flight guard; CancellableBox.resumeOnce clears its refs; Retry now flips to .loading. Error-overlay hit-testing is moot now that Retry exists.

Deferred to the upcoming menu-screen UI revamp: this PR's scope is replacing the static mock menu with the BFF integration, so these polish items will be handled in that follow-up PR:

- Menu.title → screen title
- PlatformSelectionView debug strings → L10n
- Injected UserDefaults suite in the BFFPlatform* / PlatformSelectionViewModel tests



https://github.com/user-attachments/assets/244ead8f-527b-4145-bef8-775dbdf65ef2

Prathimaboggu:
Thanks @khoinguyen-mindera — re-reviewed at `af22900`. Verified all seven carry-overs are properly fixed: route-prefix mapping (`segments[1]`, bare `/collections` dropped, tests added), `includeSubItems`/`includeMedia` gone, `hasSubCategories` now matched first, single-flight `isFetching` guard covers stale-success, `CancellableBox.resumeOnce` clears both refs, retry flips to `.loading`. The deferrals are all reasonable.

Two blockers before merge, though:

### 1. The BFF removed `platform` — `mainMenu` can't succeed

Alfie-BFF `main` at `284cfdf` (PR #17 / AF-59, merged 27 Jul) made platform config-driven and deleted the argument from every query and mutation:

```graphql
mainMenu(handle: String! = "main-menu"): Menu!
productList(after: String, collectionHandle: String!, filters: ProductFilterInput, limit: Int! = 25, sort: ProductSortEnum! = NEWEST): ProductListResponse!
```

`MenuResolver` takes only `handle`; `MenuService` reads platform from `PlatformConfigService`. Alfie-Web already adapted the same day.

This branch still sends `platform: $platform` on `mainMenu` and newly adds it to `productList`. An unknown argument is a GraphQL *validation* error, so the whole request is rejected — Shop → Categories fails against BFF `main`. Our committed `schema.graphqls` is stale, which is why codegen and CI stay green. The smoke-test video passing is the tell: that BFF was pre-AF-59.

Knock-on: the **Commerce Platform debug switcher** now toggles an argument the BFF rejects, so it should come out rather than be re-pointed. (`searchProducts`/`productDetails` already sent `platform` on `main` — that part is pre-existing, but `mainMenu` is this PR's whole payload.)

### 2. Scope

ALFMOB-387 is "rewire `getHeaderNav`, add a converter, delete the mock". The last three commits also delete the Brands tab, Store Services, segmented tabs, the `BrandsService`/`BrandsViewModel` layer, and the `alfie://brand` + `/services/store-services` deep links (which now fall through to a generic web view) — −1,821 lines since the last round. That's a product decision that needs its own ticket and design sign-off rather than riding along here.

### Nits

- `getHeaderNav`'s log still says "no recognizable collection/page/product links" — page/product are now dropped by design.
- `MainMenuIntegrationTests` uses `XCTSkipUnless(!items.isEmpty)`, so the one test that would catch `product-category` missing from an environment skips instead of failing.
- `BrandsViewSnapshotTests.swift` still references the deleted `BrandsView` (target-excluded, so it doesn't break the build, but it should go with this commit).
- `_status.md` still documents `mainMenu(handle:platform:)`.

Verified against the branch and the Alfie-BFF/Alfie-Web checkouts; I couldn't build locally (no Xcode on this machine), CI `unit-tests` is green.

khoinguyen-mindera:
Thanks @Prathimaboggu — splitting the two:

**Blocker 1 (platform)** — fixed. Re-synced `schema.graphqls` from Alfie-BFF `main` (`284cfdf`); the only diff was the `platform` removal. Dropped it from `mainMenu`/`productList`/`productDetails`/`searchProducts` + the client, and removed the Commerce Platform debug switcher. Codegen now validates against the real contract, so the drift can't hide behind a stale schema.

**Scope** — the menu **design** ([Figma – Menu Sheet](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=166-41101&m=dev)) is a plain category drill-down with no Brands tab, no Services tab, and no segmented control — bottom nav is Home / Store / Wishlist / Bag / Account. So the tab/Brands/Services removal here is implementing that design, not an unreviewed product change. Given this PR swaps the Shop screen's data source to the BFF category menu, I removed the now-superseded segmented-tab / Brands / Store-Services code rather than leave it wired up and unreachable behind the new implementation — carrying that as dead code and working around it seemed worse. The screen's **visual revamp** (search bar, styling per the design) is the separate follow-up ticket. Deep links to the removed surfaces now fall back to a web view as a consequence.

Nits: dropped the orphaned `BrandsViewSnapshotTests`, reworded the `getHeaderNav` log, hardened `MainMenuIntegrationTests` (`XCTSkipUnless` → hard assert), updated `_status.md`.

Prathimaboggu (commented):
Thanks @khoinguyen-mindera — re-reviewed at `3cad640`. Both blockers are genuinely resolved.

### 1. `platform` removal — verified clean

Diffed the committed `schema.graphqls` against Alfie-BFF `origin/main:src/schema.gql` (`284cfdf`): byte-identical apart from a trailing newline. `platform` is gone from all four operations, the regenerated Apollo types, and `BFFClientService` — the only remaining hit repo-wide is the unrelated `platformId` schema field. The Commerce Platform switcher is fully out (`BFFPlatform.swift`, `PlatformSelectionView`/`ViewModel`, the `DebugMenuView` row, both test files) with no dangling references. Agreed that codegen validating against the real contract is the right fix — this class of drift can't hide behind a stale schema again.

### 2. Scope — withdrawing it

Checked the Figma node you linked (`166:41101`): three-level plain category drill-down, no Brands tab, no Services tab, no segmented control, bottom nav Home / Store / Wishlist / Bag / Account. The removal is implementing that design, so this isn't an unreviewed product change — my objection doesn't stand. Worth a line in the ticket that `alfie://brand` and `/services/store-services` now fall through to a web view, so QA isn't surprised by it.

Also confirmed fixed: `BrandsViewSnapshotTests` dropped, log reworded, `MainMenuIntegrationTests` hard-asserts, `_status.md` updated. CI `unit-tests` green.

### One real leftover from the deletions

`AlfieUITests.swift:53-62` — `testAddToBagFullFlow`, the repo's only end-to-end add-to-bag journey, still runs Home → Shop → `segmented-option-brands` → `AccessibilityID.Brands.item`. Both are deleted here, and `continueAfterFailure = false`, so the test can only fail at the segment tap. `AlfieUITests` isn't in `Alfie.xctestplan`, which is why CI stays green. It should be repointed at Categories → leaf → PLP (which is arguably a better journey for it now anyway).

### Smaller

- `ShopViewSnapshotTests.swift` is the same orphan class as the `BrandsViewSnapshotTests` you dropped — still `ShopView<MockCategoriesViewModel, MockBrandsViewModel, MockWebViewModel>` with `initialTab: .brands` / `.services`. Target-excluded so the build is fine, but it's dead on arrival. `project.pbxproj` also still lists `Snapshots/BrandsViewSnapshotTests.swift` in `membershipExceptions` for a file that no longer exists.
- The `.storeServices` remote-config flag is now orphaned: its only UI consumer (the Services tab) is gone, but `AppDelegate.swift:83` still reboots the app when it flips, and it's still in the debug Feature Toggle list — so toggling it reboots users for no visible change.
- `AccessibilityID.Brands` has no producer left; its only consumer is the broken UI test above.
- The converter still ignores the host on the bare-single-segment path: `https://www.instagram.com/alfie` → path `/alfie` → `.listing` `/alfie` → empty PLP. The doc comment says external links are dropped, which holds for `/pages/*` etc. but not this shape. Low risk now that the handle is a curated `product-category` menu — a `components.host == nil` guard would make the code match the comment, and no test pins the foreign-host case either way.
- `plan.md` still documents `mainMenu(handle:platform:)` and `BFFPlatform.predefined` — only `_status.md` was refreshed.
- The smoke-test video was recorded at `af22900`, against a BFF that still accepted `platform`. Both the request shape and the menu handle have changed since, so one more Categories → PLP pass against a BFF at `main` would confirm `product-category` exists in that environment.

Happy for this to merge once the UI test is repointed; the rest can ride the follow-up.

Prathimaboggu (approved):

PR #98, opened 2026-07-21
ALFMOB-439: Modern Design Rollout — Home

khoinguyen-mindera:
## ALFMOB-439 — Modern Design Rollout — Home

Restyles the Home screen toward the modern Figma design ([Homepage](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=1-10082)), part of the app-wide redesign epic [ALFMOB-427](https://mindera.atlassian.net/browse/ALFMOB-427). Ticket: [ALFMOB-439](https://mindera.atlassian.net/browse/ALFMOB-439).

> **Base branch:** targets `main`. The app-shell/tab-bar work this builds on (ALFMOB-438, #97) has since been merged, so the branch was rebased onto `main` and this PR is scoped to the Home redesign.

## Scope

The BFF isn't ready yet, so this ships the UI with **mock data**, structured so a future GraphQL query drops in behind the same protocol. Per product, this ticket covers the **first hero card** + the **navigation restyle** (the remaining Figma sections — category/product/brand rails, chips — are follow-ups).

### Changes
- **Navigation:** centered **MINDERA / ALFIE** branded logo (reuses the `splash-wordmark` asset from the ALFMOB-437 splash); dropped the trailing Account + Debug toolbar buttons to match the logo-only Figma header. No divider.
- **Hero banner:** new full-bleed `HomeHeroCarouselView` / `HomeHeroBannerView` — image + bottom scrim + `display/medium` title + underlined "Explore Collection" CTA + bottom-left page dots. Driven by `HomeHeroBanner.placeholders` (mock, 3 slides) exposed via `HomeViewModelProtocol.heroBanners`.
- **Search bar:** pinned above the scrolling content; `.soft` theme now uses the 4pt (`radiusSoft`) corner radius + Figma padding; placeholder → "What are you looking for?".
- **Account:** new **Settings** row that opens the settings/debug page (relocated here since it was removed from the Home toolbar; Account is reachable via its tab). Plumbs `apiEndpointService` into `MyAccountDependencyContainer` at the composition root; `MyAccount` now depends on `DebugMenu`.
- Styling uses SharedUI semantic tokens throughout (`Theme.*`, `theme.font/spacing/radius`); no hardcoded colours/spacing/typography.

### Notes / follow-ups
- The "Settings" row opens the existing **debug menu** as a stopgap ("simple settings item for now") — a real settings screen is a future ticket.
[...]

PR #97, opened 2026-07-17
ALFMOB-438: Restyle app shell & tab bar to modern Figma design

khoinguyen-mindera:
## ALFMOB-438 — App shell, tab bar & navigation chrome (AppFeature)

Restyles the bottom tab bar to the modern Figma "Bottom Navigation" design and, per follow-up review, brings the tab set in line with the design.

🎫 https://mindera.atlassian.net/browse/ALFMOB-438

### Scope note
The original ticket was scoped visual-only (no tab-set/routing changes). During review this was **intentionally expanded** to also add the Account tab and align the Store tab, at the assignee's request. The AC deltas below are deliberate, not oversights.

### Changes
**Tab bar restyle**
- Migrated tab bar colours/typography from raw `Primitives.*` to the semantic `Theme.*` / `theme.font.label` layers.
- Selection treatment now matches Figma: removed the matched-geometry underline; selected tab = **bold label** (`label.smallBold`) + darker `Theme.contentContentPrimary`.
- Filled icon on selection (`home-fill` / `bag-fill` / `wishlist-fill` / `account-fill`) via `Tab.icon(isSelected:)`.
- Divider tinted `Theme.borderSoft`; background `Theme.surfaceBackgroundPrimary`.

**Tab set (expanded scope)**
- Added **Account** as the 5th tab (Home · Store · Wishlist · Bag · Account), reusing the existing `MyAccountFlow`.
- Renamed the **Shop** tab label to **Store** to match Figma.
- Store tab now uses the Figma glyph (the `menu` list+search icon) instead of `storefront`.
- Account → Wishlist row routes to the dedicated **Wishlist tab** (no half-wired embedded surface).

**Housekeeping**
- Tab accessibility identifiers centralised in `AccessibilityID.TabBar` (values unchanged; UI tests updated to reference them).

### Tests
- `TabBarItemStyleTests` — selected/unselected icon + label colours differ; selected label uses a heavier weight.
- `TabIconTests` — fill-variant mapping (tabs with a `-fill` swap on selection; Store has none).

### Acceptance criteria
- ✅ Tab bar matches Figma (bold selected label, filled selected icon, no underline, divider).
- ✅ Consumes `SharedUI` semantic tokens — no raw `Primitives.Colours.*` at colour sites.
- ✅ Appearance covered by passing unit tests.
- ✅ No regression in existing navigation/tab selection/deep-link routing (verified via build + unchanged routing wiring).
[...]

PR #96, opened 2026-07-15
ALFMOB-437: Modern Design Rollout — Startup / Splash screen

khoinguyen-mindera:
## ALFMOB-437 — Modern Design Rollout: Startup / Splash screen

Restyles the app startup screen to the modern Figma design, adds the design-system **Loading Spinner** component, and (for launch→in-app consistency) updates the native launch screen. Visual-only; no behaviour, timing, or boot-sequence change. No feature flag (app unreleased). Dark mode deferred.

Jira: https://mindera.atlassian.net/browse/ALFMOB-437
Figma: [Loading screen (iOS)](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=672-80063) · [Loading Spinner (DS)](https://www.figma.com/design/PWVgEoKrIw9Hv7QlOCcUoq/Alfie---Design-System?node-id=6619-47520)

### What changed
- **In-app startup screen** (`AppFeature`): new `SplashView` — the **MINDERA/ALFIE** wordmark centred (vertically aligned with the launch screen, no jump on hand-off) with the loading spinner hanging below it via `.overlay`. Replaces the animated GIF `ThemedLoaderView` for the `.loading` state; background bleeds full-screen.
- **DS Loading Spinner** (`LoadingSpinner`, SharedUI): the Figma spinner artwork (`ThemedImage.loadingSpinner`, an angular "comet" arc PNG) rotated continuously, sizes **S/M/L (24/32/48)** via a `Size` config. Splash uses `.small`. (Rendered from the exported image because the design's conic gradient can't be carried into Xcode by SVG.)
- **Wordmark asset**: `ThemedImage.splashLogo`, shared by both screens.
- **Native launch screen** (`LaunchScreen.storyboard`): new MINDERA/ALFIE `LaunchLogo`, resized to the lockup aspect (160×49), `preserves-vector-representation` restored.
- `AccessibilityID.Splash.screen`; `AccessibilityIdentifiers` wired into the `AppFeature` target.

### Acceptance criteria
- ✅ Splash matches the approved Figma.
- ✅ Consumes SharedUI components + tokens — no bespoke hardcoded colours/spacing/typography.
- ⚠️ **Appearance test deferred.** The snapshot suite is disabled repo-wide; a render smoke test added no real signal for a logic-free view, so it was removed. Appearance verification (splash + DS spinner) is deferred to a **separate snapshot-suite-revival ticket**. Verified manually in the simulator for this PR.
[...]

PR #100, opened 2026-07-24
test: snapshot testing harness with splash + home screen tests

khoinguyen-mindera:
## Ticket
N/A — test infrastructure

## Summary
- Stood up a working snapshot-testing harness. The suite was previously inert: every test was excluded from the `AlfieTests` target via `membershipExceptions`, and **no reference images had ever been recorded**
- Moved the shared helpers (`embededInContainer`, `defaultImage`) into `TestUtils`, **out of the production `Utils` module** — `TestUtils` no longer depends on `Utils`, and the snapshot container pins `displayScale` via the strategy's traits instead of mutating `UIScreen.main`. `defaultImage(precision:perceptualPrecision:)` is parameterised for per-test overrides (default `precision: 1.0`)
- Snapshot tests now live in AlfieKit module test targets, where SPM makes every file an automatic target member — they can no longer silently fall out of the build
- Pinned the verify test run to an iOS 26 simulator; the previous generic destination could resolve to an older iOS and assert against wrong-OS rendering. When no iOS 26 iPhone exists, it falls back to the newest available iPhone and **skips the snapshot classes** (discovered as any file calling `assertSnapshot`), so a missing runtime costs snapshot coverage rather than blocking every other test
- Added a fast grep guard that fails the run if a snapshot test is committed in record mode (covers `record: true/.all/.failed`, `withSnapshotTesting(record:)` and an `isRecording` flag; scans both test roots)
- Removed the dead `membershipExceptions` set and the empty `AlfieTests/Snapshots/` folder
- **Removed the six pre-existing snapshot suites** (Search, RecentSearches, Categories, Shop, ProductDetails, ProductDetailsColorSheet) — those screens are mid Modern Design Rollout (#96, #97, #98), so their references would churn on every rollout PR. They also never ran (excluded from the target, no references committed). (A seventh, Brands, was already removed on `main` in #99.)
- **Added `SplashViewSnapshotTests`** (AppFeatureTests) and **`HomeViewSnapshotTests`** (HomeTests, with/without hero banners), each with committed reference images
- **Enabled 4 test targets that were missing from `Alfie.xctestplan`** — `HomeTests`, `ModelTests`, `MyAccountTests`, `UtilsTests` — so **23** pre-existing test functions that never executed under `verify.sh` now do (all pass)
- Documented the harness, the device/precision policy, the animated-content pitfall, and the test-plan requirement in `Docs/SnapshotTesting.md`

### Notes on coverage
- The splash test covers the wordmark, placement and background — **not** the spinner, which rotates off wall-clock time. The suite default is `precision: 1.0` (a static screen can't pass with an element missing), so the splash test lowers precision to `0.9`; at `1.0` the spinner flakes every run, and at `0.9` it passes even with the spinner removed — so the spinner is better covered by a unit test.
- The `withHeroBanners` test uses a flat-colour placeholder asset instead of the photographic hero JPGs, keeping the reference small (~279 KB vs 2.3 MB) and its diffs meaningful while still covering carousel layout. Both home tests are fully deterministic (local bundle assets).


Prathimaboggu:
Reviewed at `4b110f0`.

## Overview

Turns a nominally-present-but-inert snapshot suite into a real one. Three things happen at once: the harness moves from the `AlfieTests` app target (where all 7 files were excluded via `membershipExceptions` and no references had ever been recorded) into SPM module test targets; the pre-existing screen tests are deleted and replaced with splash + home tests that do have committed references; and `test-for-verification.sh` gains a pinned simulator plus a record-mode guard. Four test targets missing from the test plan are also enabled.

The diagnosis is correct and well-evidenced. Moving to SPM test targets genuinely fixes the root cause — no `membershipExceptions` means no silent exclusion. The measured note that the splash test passes with `LoadingSpinner` deleted is the kind of finding most people don't go looking for, and documenting it rather than papering over it is the right call.

## Findings

### 1. The relocated helper reaches a private-API call that ships in the app binary

This PR adds `"Utils"` to TestUtils' dependencies so `View+Snapshots.swift` can call `embededInContainer(size:scale:)`. That function is `Alfie/AlfieKit/Sources/Utils/Extensions/View+Extension.swift:9`:

```swift
public func embededInContainer(size: CGSize, scale: CGFloat) -> UIView {
    UIScreen.main.setValue(scale, forKeyPath: "scale")   // KVC-writing a readonly UIKit property
```

Three problems compound:

- `Utils` is a **production** target — imported directly by `AlfieApp.swift`, `AppDelegate.swift`, `ServiceProvider.swift` and 9 other shipping modules. So a KVC write to a read-only UIKit property is in the release binary and its selector string is statically visible.
- Both `embededInContainer(size:scale:)` and `asUIView` are called **only** from the snapshot harness — I grepped every call site. Nothing in production uses them.
- The scale mutation is global and never restored, so every test running after a snapshot test in the same process sees a mutated `UIScreen.main.scale`.

Since this PR is the one consolidating the harness, this is the moment: move both functions into `TestUtils` and delete them from `Utils`. That removes the private-API call from the app, contains the global mutation to test-only code, and lets you drop the new `Utils` dependency entirely rather than adding it.

It also qualifies the doc's "model is free" claim. The three devices verified (17 Pro, 17, 16) are all @3x, so the forced scale is a no-op there. On an @2x iPhone the forced-scale path is load-bearing and untested. Worth saying "any **@3x** iPhone at iOS 26" in the policy.

### 2. The iOS pin gates the whole suite, not just the snapshot tests

`test-for-verification.sh` hard-exits when no iOS 26 iPhone is installed, and the old fallback path is deleted. That means a snapshot-testing constraint now blocks every non-snapshot assertion in the suite: a dev on the next Xcode (iOS 27 simulators only) can build but cannot run *any* unit test. Given `AGENTS.md` points agents at `verify.sh`, that failure will be hit.

`SNAPSHOT_OS_MAJOR` is a good escape hatch, but the error message doesn't mention it — it only says to install iOS 26. Two cheap improvements: name the env override in the error text, and consider falling back to the newest available iPhone with a loud warning plus `-skip-testing:` on the snapshot classes, so a missing runtime costs you snapshot coverage rather than all coverage.

### 3. The record-mode guard has gaps that matter for this library version

The grep guard matches `record:\s*true` and `isRecording\s*=\s*true`. On snapshot-testing 1.18.3, `record:` takes `SnapshotTestingConfiguration.Record?`, so all of these evade it:

- `record: .all` — the idiomatic modern spelling
- `withSnapshotTesting(record: .all) { ... }`
- `private let isRecording: Bool = true` — the type annotation breaks the `=` adjacency
- the library's own `SNAPSHOT_TESTING_RECORD` env var (ungreppable, worth a doc note)

Also the guard only scans `Alfie/AlfieKit/Tests`. `Alfie/AlfieTests` still exists (`Services/ApiEndpointServiceTests.swift`), and `TestUtils` is importable from it, so a snapshot test added there is unguarded. Adding `.all`/`.failed` to the pattern and scanning both roots is a one-line change.

### 4. A 2.3 MB reference PNG

`test_homeView_withHeroBanners.1.png` is 2,307,930 bytes — 33× the other two, because it captures photographic hero assets at 1179×2556. Every re-record adds another 2.3 MB permanently to git history, and the PR's own rationale is that references churn on design changes. Either point the mock at a flat-colour placeholder asset (you keep layout coverage, which is what the test is for), or put `__Snapshots__/**.png` behind git-lfs before this pattern spreads.

### 5. Doc and PR-body claims that don't hold

- `Docs/SnapshotTesting.md` says TestUtils "is already a dependency of every module test target." Five aren't: `BagTests`, `BFFGraphTests`, `ModelTests`, `SharedUITests`, `UtilsTests`. The "Adding a snapshot test" section says "most already do," which is right — the two contradict each other in the same file.
- PR body says "~50 existing test runs that never executed." Actual count is **25** across the four targets (Home 7, Model 12, MyAccount 5, Utils 1) — and 2 of those 7 are new in this PR, so 23 pre-existing. The doc's "25 test functions" is the accurate number; the body is 2× off.
- The body says the removed suite was "Search / CategorySelector / ProductDetails," but `BrandsViewSnapshotTests` and `ShopViewSnapshotTests` also went. If Brands isn't in the Modern Design Rollout scope (#96–#98), the stated rationale doesn't cover it — worth either migrating it now or naming it in the list.

### 6. Dead parameter in the newly-public helper

`View+Snapshots.swift:10-22` — `coordinatedView(view:)` returns its argument unchanged, so `coordinate: Bool = true` does nothing on either branch. Pre-existing, but this PR promotes it to public API of a library product — good moment to delete both.

## Smaller

- The deleted files carried a useful scenario inventory (loading/error/empty across Brands and Shop, PDP with single/many/no colours). Even though they never ran, re-adding per screen now starts from a blank page. Consider listing the scenarios in `Docs/SnapshotTesting.md` so the intent survives the deletion.
- `defaultSnapshotSize` still comments `// iPhone 15 Pro`, but the policy is now "any iPhone at iOS 26" and references were recorded on a 17 Pro (402×874 logical). The fixed size is fine — the comment now reads as a device requirement it isn't.
- Because the container is a bare fixed-size `UIView`, these snapshots have no safe-area insets, so they can't catch safe-area regressions. Worth one line in the doc's "what this covers" framing.
- The inline `/usr/bin/python3` is a new hard dependency in a script that previously used only `grep`/`sed`. Fine on any machine with CLT, but it will prompt an install if absent — worth a comment or a graceful message.
- `build-for-verification.sh` still uses the generic destination. Correct as-is (`test-without-building` consumes the generic build), just noting the two scripts now disagree on destination strategy in case that surprises someone later.

## Coverage and risk

Two screens with three references is a thin but honest starting point, and it's strictly better than seven files that never compiled. Enabling the four missing test targets is the highest-value change here and is independently valuable. Security surface is limited to the private-API point in finding 1.

Nothing structural is wrong. **Findings 1 and 2** are the ones I'd want addressed before merge; 3 and 4 are cheap and prevent the pattern from calcifying; the rest are doc corrections.

_Note: I couldn't run `verify.sh` (no Xcode on my machine), so the "all pass" claim and the reference images are unverified on my side. Everything above is from reading the diff and the surrounding code._

khoinguyen-mindera:
Thanks @Prathimaboggu — this was a genuinely useful review, all six findings were on the money. Addressed across three commits (`70a2a56`, `68e5674`, `1a25d6a`); `verify.sh` is green.

**1 — private-API KVC in production `Utils`.** Moved `embededInContainer`/`asUIView` into `TestUtils` (now `private`), deleted them + the `UIScreen.main.setValue(..., "scale")` write from `Utils`, and dropped the `TestUtils → Utils` package edge entirely. The scale is now pinned via the strategy's traits (`displayScale: 3`) instead of the global mutation. The splash + without-banners references matched byte-for-byte after the move, confirming no rendering change. Good catch that the KVC write was load-bearing for the fallback scale — the traits pin makes "model is free" genuinely true now (even a @2x device renders @3x).

**2 — the pin gated the whole suite.** Two parts: the error now names the `SNAPSHOT_OS_MAJOR` override, and (your fuller suggestion) when no iOS 26 iPhone exists the script falls back to the newest available iPhone and `-skip-testing:`s the snapshot classes — discovered dynamically as any file calling `assertSnapshot`, so no hardcoded list to maintain. A missing runtime now costs snapshot coverage, not every test. Verified both paths.

**3 — record-guard gaps.** Broadened to also catch `record: .all/.failed`, `withSnapshotTesting(record:)`, and a type-annotated `isRecording: Bool = true`, and it now scans `AlfieTests` too. (Small correction for the record: one — `record: .all` at the `assertSnapshot` call site is actually `Bool?` in 1.18.3 so it wouldn't compile; the real leaks were `withSnapshotTesting` and the env var, both covered now.)

**4 — 2.3 MB reference.** `withHeroBanners` now points at a flat-colour placeholder asset instead of the photographic JPGs: **2.3 MB → 279 KB**, layout coverage preserved.

**5 — doc/PR-body claims.** Fixed the "dependency of every module test target" claim, the removed-file list, the dangling "migration playbook" ref, and the repin-vs-env contradiction. PR body corrected too (`~50` → 25, accurate 7-file list).

**6 — dead `coordinate:` param.** Removed, along with the no-op `coordinatedView`.

Also tightened the default `precision` to `1.0` (a static screen can no longer pass with an element missing) and lowered only the splash test to `0.9`, since its spinner flakes at `1.0`.

**Deferred (happy to take in a follow-up):** the `LoadingSpinner` unit test the doc recommends; CI sharing the same OS pin as `verify.sh` (CI runs bare `scan` today); and the remaining smaller nits (safe-area doc line, the new `python3` dependency note, the `build-for-verification.sh` destination asymmetry).

Prathimaboggu (commented):
Re-reviewed at `1a25d6a`. All six findings are properly fixed — I checked the code rather than the summary, and in a couple of places the library source too.

## Confirmed fixed

**1 — private-API KVC.** Gone, not just relocated: no `UIScreen.main.setValue` anywhere in the repo, both functions deleted from `Utils`, the `TestUtils → Utils` package edge dropped, no residual callers of `asUIView`/`defaultSnapshotScale`.

I also verified the replacement mechanism, since the byte-for-byte match was measured on a @3x device where old and new both yield 3. In 1.18.3 `renderer(bounds:for:)` builds `UIGraphicsImageRendererFormat(for: traits)`, and the diffing scale reads `traits.displayScale` as well (`Snapshotting/UIView.swift:34`) — so render and compare agree at @3x on any host. The traits pin makes "model is free" true by construction rather than by accident, which is a better outcome than the KVC hack it replaces.

**2** — fallback + dynamic `-skip-testing:` discovery behaves as described; the error text names `SNAPSHOT_OS_MAJOR`. **3** — guard broadened, `--include` before the path, both roots scanned. **4** — 278,854 bytes on disk. **5** — 25 confirmed exactly (Home 7 + Model 12 + MyAccount 5 + Utils 1), 7-file list accurate, all 18 targets enabled in the plan, TestUtils claim now correctly hedged (13 of 18 targets depend on it). **6** — dead param and `coordinatedView` gone.

And you're right about `record: .all` — I was wrong. At 1.18.3 `assertSnapshot(…, record:)` is `Bool?` (`AssertSnapshot.swift:106`); `SnapshotTestingConfiguration.Record` only appears on `withSnapshotTesting`. The leak you covered was the real one.

## Worth folding in

**CI is the deferred item that will actually bite, and it's a one-liner.** `Fastfile` calls bare `scan` and `.env.default` sets `SCAN_TESTPLAN` but no device. The job is `runs-on: macos-26`, so the iOS 26 pin holds today *only* because that runner image happens to carry iOS 26 simulators. Whenever someone bumps to `macos-27` for an unrelated reason, CI asserts iOS-26 references against iOS 27 with neither the fallback nor the skip `verify.sh` has — and it'll read as a rendering regression rather than config drift. `.env.default` already centralizes scan config, so `SCAN_DEVICES=...` there is the whole fix.

**The two root lists disagree.** `test-for-verification.sh:81` scans both test roots for the record guard; line 151 discovers snapshot classes under `Alfie/AlfieKit/Tests` only. A snapshot test in `AlfieTests` would be guarded against record mode but not skipped on the fallback path, so it'd assert on the wrong OS and fail — and the `sed` couldn't derive a target from that path shape anyway. Reuse `SNAPSHOT_TEST_ROOTS`.

**A test fixture landed in a production asset catalog.** `hero-placeholder.imageset` sits in `Sources/Home/Resources/HomeAssets.xcassets/`, referenced only from `HomeViewSnapshotTests.swift:27`, so it ships in the app. It's forced — `HomeHeroBannerView.swift:35` resolves `bundle: .module` against Home's own bundle — and it's 74 bytes, so this is naming/hygiene, not cost: worth marking it test-only so nobody later mistakes it for a real placeholder.

## Smaller (follow-up is fine)

- `--filter HomeTests/HomeViewSnapshotTests` on the fallback path emits both `-only-testing:` and `-skip-testing:` for the same identifier; skip wins, so you get a green run that executed zero tests. A warning when the filter intersects the skip list would avoid a confusing green.
- Class discovery takes only the first `class` per file, so a second test class in one file silently wouldn't be skipped. Fine today, a trap later.
- `SNAPSHOT_TESTING_RECORD` still isn't mentioned anywhere in the repo. Low stakes since record mode always fails and so can't produce a false green, but it isn't covered as claimed — one doc line.
- `withSnapshotTesting(record:` matches unconditionally, so a legitimate `record: .never` trips the guard.
- "25 **existing** test functions" is 23 existing + 2 new in this PR.

## Verdict

Both pre-merge blockers are resolved; nothing remaining blocks merge. I'd take the CI pin here since it's one line, and the rest can follow.

Same caveat as last time: no Xcode locally, so I couldn't run `verify.sh` or re-render the references. CI `unit-tests` is green in 10m53s on macos-26, which does exercise them at iOS 26. The "@2x renders @3x" claim is correct per the library source but stays empirically untested with no @2x simulator in play.

khoinguyen-mindera:
Thanks for the thorough re-review @Prathimaboggu — and for verifying against the library source. Folded in the two you'd take now, in `bc49639`:

**CI pin.** `.env.default` now sets `SCAN_DEVICE=iPhone 17 Pro (26.2)` — the simulator the `macos-26` runner actually resolved on the last green run, so it's known-good rather than a guess. Bare `scan` was picking the runner's newest runtime, so this is the fix for the exact drift you described: a future `macos-27` bump would otherwise assert iOS-26 references on iOS 27 as a phantom regression. Commented that the string tracks the runner's iOS 26 sim and needs a bump if that image changes.

**Root-list disagreement.** Good catch — that was a gap I introduced. The fallback discovery now scans `SNAPSHOT_TEST_ROOTS` (both roots, same as the record guard) and derives the target from either path shape (`.../AlfieKit/Tests/<Target>/...` or `.../AlfieTests/...` → `AlfieTests`), so a snapshot test added under `AlfieTests` would be skipped on the fallback path, not just guarded against record mode.

**Test fixture in the production catalog.** Acknowledged — it does ship. It has to live in `HomeAssets` because `HomeHeroBannerView` resolves `bundle: .module` against Home's own bundle, so a test target can't hold it. At 74 bytes it's hygiene not cost, so I've left the asset in place for now; happy to rename it test-only if the team prefers that signal.

**Smaller items** — all fair, deferring as you suggested: the filter∩skip "green run with zero tests" warning, first-`class`-per-file discovery limit, the `SNAPSHOT_TESTING_RECORD` doc line, and the `record: .never` false-positive in the guard. On "25 existing" — you're right it's 23 pre-existing + 2 new in this PR; I'll tighten that wording.

CI `unit-tests` is green on `macos-26` (exercises the suite at iOS 26); `verify.sh` passes locally.

Prathimaboggu (commented):
Third pass, at `bc49639`. @khoinguyen-mindera — both items I asked for last round landed correctly and the two first-round blockers remain fixed. **Nothing blocks.** Two of the three items below are in the new commit; the third is a hole the (correct) hero-asset swap opened.

## What I verified — and the caveat

Still no Xcode on this machine (Command Line Tools only, macOS SDKs only, no iOS SDK, no `simctl`, no `swiftlint`), so **`verify.sh` was not run and this is a static review**. I did not build or execute a single test.

Instead: I read the repo's own CI run for this head (`30467325880`, `unit-tests` green on `macos-26`), which is the first run to exercise the new `SCAN_DEVICE` pin — the destination resolved to `iPhone 17 Pro, OS:26.2` with no fallback line anywhere in the log, all three snapshot tests passed, and all 17 SPM bundles plus `AlfieTests` reported green, including the four newly enabled targets. I decoded the three reference PNGs and measured them against the source: all are 393×852 at displayScale 3, and every value they pin — the search field's insets and height, the carousel's height against `.aspectRatio(375.0/500.0)`, the scrim's `scrimEndLocation`, the page dots' size, inset and spacing — matches the tokens exactly. `hero-placeholder.png` is 8×8 and a single flat colour, so the upscale is interpolation-free and, given `HomeHeroBannerView` renders into an explicit `size` frame with `.scaledToFill().clipped()`, the swap cannot shift layout — "layout coverage preserved" holds by construction, not by luck. Reference weight is now 424 KB total against 2.3 MB before, so that one's resolved. The first-round fixes are still clean at head: no `UIScreen.main.setValue` anywhere, both helpers `private` in `TestUtils`, no production target depending on `TestUtils`, `Package.swift` untouched. Read from Figma dev mode, the splash reference matches node `672:80063` on every property — background token, wordmark 160×49, spinner 24, and the 16pt `spacing-s` gap — except the vertical anchor, which is #96's deliberate "aligns with launch screen" choice.

## Blocking

Nothing.

## Should fix

### 1. The new CI pin fails open, so a green `unit-tests` doesn't prove it was honoured

`fastlane/.env.default:43`

`SCAN_DEVICE=iPhone 17 Pro (26.2)` is the right env var and the right spelling (fastlane documents `'iPhone SE (2nd generation) (14.5)'`), but `scan` treats an unresolvable device as advisory. In `scan/lib/scan/detect_values.rb` at 2.234.0 — the version in `Gemfile.lock` — the no-match path is:

```
UI.error("Ignoring '#{device_string}', couldn't find matching simulator")
…
UI.error("Couldn't find any matching simulators for '#{devices}' - falling back to default simulator")
```

`UI.error`, not `UI.user_error!`. The hard-failure path is gated on `ensure_devices_found`, which defaults to `false` — your own run prints it in the options table: `ensure_devices_found | false`.

Concrete failure: the `macos-26` image is refreshed (the workflow pins no Xcode version), its bundled runtime moves off 26.2 or the `iPhone 17 Pro` name goes, fastlane logs two red lines into a ~5,400-line log, and the suite runs on the newest runtime instead. That is exactly the pre-PR behaviour, now with a comment above it asserting the opposite — and the first symptom is a snapshot pixel diff that reads as a rendering regression, the misdiagnosis this commit exists to prevent.

One line, and it has to go in the `Fastfile` because `ensure_devices_found` has no `env_name`, so `.env.default` can't reach it:

```ruby
lane :test do
  scan(ensure_devices_found: true)
end
```

Worth noting the two mechanisms now disagree in strictness: `verify.sh` pins the iOS **major** and falls back loudly with snapshots skipped; CI pins model + **minor** and falls back silently with snapshots still asserting. `Docs/SnapshotTesting.md` documents only the first, so a reader of the policy won't know CI has a second one.

### 2. The record guard makes the documented record loop unreachable through the repo's own entry points

`Alfie/scripts/test-for-verification.sh:81-92` vs `Docs/SnapshotTesting.md:60-68`

The guard runs before simulator resolution and before `xcodebuild`, and `exit 1`s with no override. `verify.sh:55` forwards `--filter` straight into this script, and `AGENTS.md:62` makes `verify.sh` the mandated entry point after every change. So following `Docs/SnapshotTesting.md` step 1 ("Set `isRecording = true`") and step 2 ("Run the test") yields:

```
❌ ERROR: A snapshot test is committed in record mode (see matches above)
Set it back to false and re-run so the test asserts against the committed reference.
```

An instruction to undo step 1, with no hint that recording is possible at all. `Docs/Testing.md:53` ("Record mode: Set `isRecording = true` temporarily") names no runner either. Recording is only achievable by bypassing the scripts entirely — Xcode or raw `xcodebuild` — which is documented nowhere. For a script whose header is "Test script for AI agents", that's a dead end an agent can't reason its way out of.

Cheapest fix, keeping the guard's real job (catching *leftover* record mode) intact:

```bash
if [ "${SNAPSHOT_ALLOW_RECORD:-0}" != "1" ]; then
    …existing guard…
fi
```

plus naming `SNAPSHOT_ALLOW_RECORD=1` in the error text and in the doc's step 2.

### 3. After the placeholder swap, nothing checks that the three production hero assets resolve

`Alfie/AlfieKit/Tests/HomeTests/HomeViewSnapshotTests.swift:26-28`

The swap was the right call and I asked for it — but it left a hole worth three lines to close. The test rewrites every banner's `imageName` to `hero-placeholder`, so the only test that renders the carousel never touches `hero-1/2/3`. Nothing else covers them: `HomeViewModelTests.swift:39` asserts the *model strings* `["hero-1", "hero-2", "hero-3"]`, not that those names resolve in the catalog. And the two sides are only loosely coupled — the names live in `Model` (`HomeHeroBanner.swift:20-36`), the catalog lives in `Home`, joined at runtime by a string lookup in `HomeHeroBannerView.swift:35`:

```swift
Image(banner.imageName, bundle: .module)
```

So renaming or dropping `hero-2.imageset` ships a hero page that draws the scrim and the title over nothing, and the whole suite stays green. `Home` declares `resources: [.process("Resources/HomeAssets.xcassets")]`, so `@testable import Home` gives you its generated `Bundle.module`:

```swift
func test_placeholderHeroAssets_resolveInHomeBundle() {
    for banner in HomeHeroBanner.placeholders {
        XCTAssertNotNil(UIImage(named: banner.imageName, in: .module, with: nil), banner.imageName)
    }
}
```

(If the generated accessor isn't reachable under `@testable`, a one-line internal helper in `Home` does the same job.)

And to answer your open question on the asset name: yes, worth renaming. `hero-placeholder` sitting in a production catalog next to `hero-1/2/3` reads as *the* production placeholder for hero images — a plausible thing for someone to wire into a loading state later. `test-hero-placeholder` costs nothing and removes the ambiguity. The 74 bytes aren't the issue; the name is.

## Verification gaps

- **No local build or test run.** Everything above is diff reading, token arithmetic, PNG decoding, call-site greps, Figma dev mode, and the repo's own CI log. I have not re-rendered a reference and can't.
- **The pin's effectiveness isn't observable from the check status** — see should-fix 1. Until `ensure_devices_found` is on, "CI is green" and "CI ran on iOS 26" are independent facts.
- **Description drift.** `main` deleted `AlfieTests/Snapshots/BrandsViewSnapshotTests.swift` itself in #99 (`3d2efac`), so post-merge this PR removes **six** suites, not seven — the body and `Docs/SnapshotTesting.md:83-85` both still say seven and list Brands. Also still outstanding from last round: "**25** existing test functions" is 23 existing + 2 new in this PR (Home 7 = 5 + 2, Model 12, MyAccount 5, Utils 1).

khoinguyen-mindera:
Thanks again @Prathimaboggu — thorough as ever, and all three should-fix items were spot on. Fixed in `d5f9c68`; `verify.sh --skip-integration` is green (including the new asset test).

**1 — CI pin fails open.** Confirmed against the pinned `scan` (2.234.0): `options.rb` defaults `ensure_devices_found: false` and `detect_values.rb:249-260` only hard-fails when it's on, otherwise logging `UI.error` and falling back to the default sim. Since the option has no `env_name`, `.env.default` can't reach it, so the fix goes in the `Fastfile`: `scan(ensure_devices_found: true)`. A green run now proves the pinned simulator was actually used rather than silently swapped.

**2 — record loop unreachable through the scripts.** Wrapped the guard in a `SNAPSHOT_ALLOW_RECORD=1` bypass, named it in the error text, and documented the record command in `Docs/SnapshotTesting.md` step 2. The guard still fails by default on an accidentally-committed record mode — the bypass only lifts it for a deliberate re-record run.

**3 — nothing checked the real hero assets resolve.** Added `test_heroBannerAssets_resolveInHomeBundle`, which asserts `hero-1/2/3` resolve in Home's bundle, so dropping/renaming one fails a test instead of silently shipping a blank hero. You were right that `.module` in the test target resolves to the *test* bundle, so I added a `BundleToken` in Home (mirroring `SharedUI+Bundle.swift`) for the test to reach Home's bundle. And renamed the fixture `hero-placeholder` → `test-hero-placeholder` — content is byte-identical so the reference PNG still matches, no re-record.

**Still outstanding (description-only, will tidy separately):** the "seven → six suites" drift now that #99 already deleted `BrandsViewSnapshotTests`, and the "25 existing" wording (23 existing + 2 new). Neither affects behaviour.

Prathimaboggu (commented):
@khoinguyen-mindera — fourth pass, reviewing `8e83367..2cb6793` (the merge of `main`, plus `d5f9c68` and `2cb6793`). Static review: Command Line Tools only here, so no local build, tests or lint — this is diff reading, script tracing and the repo's own CI log for head (`30827072095`, `unit-tests` green on `macos-26`). All three round-3 items landed correctly and **nothing blocks**. Two should-fix items below.

`Home+Bundle.swift` mirroring the existing `SharedUI+Bundle.swift` accessor, and the fixture rename landing as a pure git rename so the reference PNG still matched without a re-record, were both the right calls.

## Should fix

### 1. The new asset guard sits inside the one class the fallback path skips

`Alfie/AlfieKit/Tests/HomeTests/HomeViewSnapshotTests.swift:46`

`test_heroBannerAssets_resolveInHomeBundle` has no reference image and no OS dependency, but it was added to a class that has both. `Alfie/scripts/test-for-verification.sh:160` discovers snapshot files with `grep -rlE 'assertSnapshot\('`, and `:158-159` emits a **class-level** skip:

```
-skip-testing:HomeTests/HomeViewSnapshotTests
```

So on any machine without an iOS 26 iPhone — the fallback path added in `68e5674` — the asset test is skipped along with the two snapshot tests, and `verify.sh` still prints `✅ TESTS PASSED`. The regression the test exists to catch (rename or drop `hero-2.imageset` → the hero page draws its scrim and title over nothing) still ships silently there, and that's the path a dev on a newer Xcode will be on.

The fix is a move, not a rewrite: put it in its own class, or in `Alfie/AlfieKit/Tests/HomeTests/HomeViewModelTests.swift` — already in `HomeTests`, and its `import SharedUI` is non-`@testable`, so SharedUI's internal `BundleToken` stays invisible and there's no ambiguity with Home's. Either keeps the file out of the discovery grep. CI is unaffected; `scan` never emits the skip list.

### 2. The newly documented record command's `--filter` won't resolve

`Docs/SnapshotTesting.md:64` vs `Alfie/scripts/test-for-verification.sh:73-75`

```
SNAPSHOT_ALLOW_RECORD=1 ./Alfie/scripts/verify.sh --skip-integration --filter <TestClass>
```

`--filter` is interpolated verbatim into `FILTER_ARG="-only-testing:$TEST_FILTER"`. xcodebuild reads the first component of an `-only-testing:` identifier as the *test target*, so substituting `HomeViewSnapshotTests` names a target that doesn't exist — it either errors out or matches nothing and exits 0 having written no PNG. Either way step 3 ("inspect the produced PNG") has nothing to inspect. Should read:

```
--filter HomeTests/HomeViewSnapshotTests
```

`PLAUSIBLE` rather than confirmed — no Xcode on this machine, so I couldn't run xcodebuild to establish which of the two failure modes it takes. The script's own `--help` example (`'LocalizationTests'`, a class in `SharedUITests`) has the same shape and is pre-existing; the doc line is new in `d5f9c68`.

Corollary, one line: with the correct `HomeTests/HomeViewSnapshotTests` form, the previously-deferred filter∩skip collision now sits on a **documented** path — on a non-iOS-26 machine the record run emits `-only-testing:` and `-skip-testing:` for the same identifier, skip wins, and you get a green run that recorded nothing. Worth pulling that warning forward, or noting in the doc that recording needs the pinned runtime.

Prathimaboggu (approved):

Issue #31, opened 2025-02-14
Build phase fails when `SRCROOT` contains a space character


ashikase:
The script for the build phase `Copy GoogleService-Info.plist` fails when `SRCROOT` contains a space character, e.g., `/path/to/folder of/Alfie-iOS`.

https://github.com/Mindera/Alfie-iOS/blob/3b86301e333b115373885b47d116c7ecc882d817/Alfie/Alfie.xcodeproj/project.pbxproj#L356

The last line of the script:

```shell
cp $SOURCE_CONFIG_PATH $DESTINATION_CONFIG_PATH
```

...should be:

```shell
cp "$SOURCE_CONFIG_PATH" "$DESTINATION_CONFIG_PATH"
```
(Technically, for this issue, quotes are only needed around the first variable.)

JoaoPinhoMinder:
@ashikase Thanks for sharing 🙏

PR #101, opened 2026-07-24
fix: assert a seed-independent variant invariant instead of SKU

khoinguyen-mindera:
## Ticket
N/A — test fix

## Summary
- `ProductDetailsIntegrationTests.test_getProduct_variantsSurface` fails on `main` against a local BFF, because it asserts `product.defaultVariant.sku.isEmpty == false`
- **Every variant in the seed Shopify store has an empty SKU** — verified by querying the BFF directly across the whole `women` collection (`sandals-lv`, `black-winter-coat`, `capucines-bb`, `high-waist-jeans`, `mini-v-neck-pleat-dress` — all return `sku: ""`). Most are "Default Title" variants, so SKU was never populated
- The variants themselves surface correctly, so the first assertion was always fine; only the SKU one was unsatisfiable
- Replaced it with a seed-independent invariant: the default variant must be one of the product's variants. `Product.Variant` has no `id`, so that was not an option
- `./Alfie/scripts/verify.sh` now reports **✅ FULL VERIFICATION PASSED (build + unit + integration)**; integration is 9/9

## Screenshot

### DIFF
--- a/Alfie/AlfieKit/Tests/BFFIntegrationTests/ProductDetailsIntegrationTests.swift
+++ b/Alfie/AlfieKit/Tests/BFFIntegrationTests/ProductDetailsIntegrationTests.swift
@@ -21,7 +21,9 @@ final class ProductDetailsIntegrationTests: IntegrationTestCase {
         let product = try await sut.getProduct(handle: slug)
 
         XCTAssertFalse(product.variants.isEmpty, "Product details should expose at least one variant")
-        XCTAssertFalse(product.defaultVariant.sku.isEmpty)
+        // Not asserting on `sku`: every variant in the seed store has an empty one.
+        XCTAssertTrue(product.variants.contains(product.defaultVariant),
+                      "The default variant should be one of the product's variants")
     }
 
     func test_getProduct_unknownHandle_throwsBFFRequestError() async throws {
--- a/Alfie/AlfieKit/Tests/BFFIntegrationTests/ProductDetailsIntegrationTests.swift
+++ b/Alfie/AlfieKit/Tests/BFFIntegrationTests/ProductDetailsIntegrationTests.swift
@@ -21,9 +21,11 @@ final class ProductDetailsIntegrationTests: IntegrationTestCase {
         let product = try await sut.getProduct(handle: slug)
 
         XCTAssertFalse(product.variants.isEmpty, "Product details should expose at least one variant")
-        // Not asserting on `sku`: every variant in the seed store has an empty one.
-        XCTAssertTrue(product.variants.contains(product.defaultVariant),
-                      "The default variant should be one of the product's variants")
+        // Not asserting on `sku`: the seed store's variants may carry an empty one.
+        // The default variant must surface media so the PDP carousel isn't blank
+        // (see the media fallback in ProductDetails+Converter).
+        XCTAssertFalse(product.defaultVariant.media.isEmpty,
+                       "The default variant should surface media for the PDP carousel")
     }
 
     func test_getProduct_unknownHandle_throwsBFFRequestError() async throws {


Prathimaboggu (commented):
Diagnosis is right — the SKU assertion was unsatisfiable, and `ProductDetailsConverterTests.swift:240` already asserts `defaultVariant.sku == ""` as an expected converter shape, so an integration test demanding a non-empty SKU was always wrong. Two things to fix.

**1. The replacement assertion can never fail.** In `ProductDetails+Converter.swift`: line 66 returns `variants.isEmpty ? nil : 0`, so the index is non-nil whenever there's a variant; line 20 sets `defaultVariant = domainVariants[index]`, an element of that array; line 47 assigns the same array to `product.variants`. So `variants.contains(defaultVariant)` is true by construction, and the only case that falsifies it — empty `variants` — is already asserted on the line above. (`Variant`'s synthesised `==` is reflexive here; `Money.amount` is `Int`, so no NaN edge.)

The old assertion was unsatisfiable, the new one is unfalsifiable — `test_getProduct_variantsSurface` now substantively asserts only non-emptiness, nothing about the default variant. Something a live response could actually falsify, still seed-independent:

```swift
// the resolution rule the converter documents at line 63
if product.variants.contains(where: { $0.stock > 0 }) {
    XCTAssertGreaterThan(product.defaultVariant.stock, 0)
}
// or the media fallback at line 115, which has a comment about the carousel going blank
XCTAssertFalse(product.defaultVariant.media.isEmpty)
```

**2. The empty SKUs deserve a ticket, not just a test comment.** Three places treat SKU as variant identity:

- `SelectedProduct.swift:43` — `id = "\(product.id)-\(selectedVariant.sku)"`, so with empty SKUs every variant of a product collapses to one `id` (and `SelectedProduct` is `Identifiable, Hashable`)
- `ProductDetailsViewModel.swift:270` — `variants.first { $0.sku == persistedSku }` matches the *first* variant regardless of what the user picked, so Bag/Wishlist → PDP silently switches the selection, which is the opposite of the "keeping the user's choice" comment above it
- `PersistedProductDTO.swift:48`/`:100` — persists and rehydrates by `variantSku`

`Product.swift:63` also documents `sku` as "A unique identifier for the variant", which the seed data violates. Real Shopify data usually does populate SKU and the seed's "Default Title" variants don't, so this may not bite in production — but nothing in the code defends against it either way. Worth raising rather than leaving as a benign-sounding note in a test.

Minor: the comment states a positive fact about a mutable external seed store, in a file whose header says it "stays seed-agnostic" — "may be empty / not guaranteed by the seed" ages better.

Context: this target isn't in CI's test plan (`SCAN_TESTPLAN=Alfie`, which excludes `BFFIntegrationTests`), so `verify.sh` locally is the only signal — which is how the broken assertion sat on `main`.

Nothing here makes anything worse and it does unblock `verify.sh`; I'd just swap the assertion for one with actual coverage first. No Xcode or local BFF on my side, so the 9/9 is unverified — the above is from the converter and model code.

Prathimaboggu (commented):
Re-reviewed at `020c4b2`. The assertion is genuinely falsifiable now — I checked it isn't another tautology.

`Variant.media` is `colour?.media ?? []`, and the converter reaches empty media on two live paths: a colour option present but `resolvedMedia` empty, or no colour option and `resolvedMedia` empty → `colour = nil`. Both occur when the variant carries no media *and* the product has no parseable `primaryImage`. So the test fails precisely in the blank-carousel case, and it exercises the GraphQL selection of `primaryImage`/variant `media` plus the fallback wiring end to end — real live-BFF coverage, which containment never had. Good pick. The softened comment ("may carry an empty one") covers the minor note too.

**Two things still open, neither blocking.**

**The SKU identity point has no ticket.** I checked rather than assuming: nothing in GitHub issues (only #31, from Feb 2025), and nothing relevant in Jira — `text ~ "sku"` over the last 60 days returns AF-84/85/88/89/91 (cart implementation) and ALFMOB-388 (Android PDP design discrepancies); `summary ~ "variant"` over 90 days is empty. The body still says "Ticket: N/A".

Agreed this isn't this PR's code to change — my ask was only that it exist somewhere, and it doesn't yet. The user-visible piece is `ProductDetailsViewModel.swift:270`: restoring a persisted selection by matching an empty SKU picks the first variant, so Bag/Wishlist → PDP silently switches the user's choice. `SelectedProduct.swift:43` and `PersistedProductDTO.swift:48`/`:100` are the other two sites.

**The PR body is now stale.** It still reads "Replaced it with a seed-independent invariant: the default variant must be one of the product's variants. `Product.Variant` has no `id`, so that was not an option" — that assertion was deleted in this commit. The body describes round one of two.

Happy with the code change. Same caveat as before: no Xcode or local BFF on my side, so the integration run is unverified here — the above is traced through the converter and model.

Prathimaboggu (approved):

khoinguyen-mindera:
Thanks @Prathimaboggu — good catch on both counts.

**Assertion (point 1):** swapped the unfalsifiable `variants.contains(defaultVariant)` for an unconditional, seed-independent check that the default variant surfaces media (`020c4b2`):

```swift
XCTAssertFalse(product.defaultVariant.media.isEmpty,
               "The default variant should surface media for the PDP carousel")
```

Went with `media` over the `stock` invariant on purpose: the stock version needs an `if product.variants.contains(where: { $0.stock > 0 })` guard, and a runtime condition around the only assertion means the test can pass by asserting nothing when live inventory sells out — same silent-vacuous failure mode as the original SKU check. `media` is falsifiable by construction (`fallbackMedia` is `[]` when there's no product image, so it isn't true-by-construction) and guards exactly the blank-carousel case the converter's fallback exists to prevent. Also reworded the comment to "may be empty / not guaranteed by the seed" so it stays seed-agnostic.

**Empty SKUs (point 2):** agreed this deserves a ticket rather than a test comment — the `SelectedProduct.id` collapse and the `ProductDetailsViewModel` first-match are real correctness issues independent of this PR. Raising it separately.

**Verification:** unchanged from your note — `BFFIntegrationTests` isn't in CI's plan and needs a local BFF, so the pass is local-only. Flagging so it's not read as CI-green.

PR #102, opened 2026-07-27
ALFMOB-467: Modern Design Rollout — PLP (grid + filter bar)

khoinguyen-mindera:
## ALFMOB-467 — Modern Design Rollout — PLP (grid + filter bar)

Restyles the Product Listing screen toward the modern Figma design ([PLP](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=1-2568)), part of epic [ALFMOB-427](https://mindera.atlassian.net/browse/ALFMOB-427). Slice 1 of 3 for [ALFMOB-440](https://mindera.atlassian.net/browse/ALFMOB-440). **Visual only, no behavioural change.** Targets `main` (rebased after ALFMOB-387 merged).

### Changes (10 files)

**PLP surface (`ProductListing/`)**
- `ProductListingView.swift` — spacings → `theme.spacing.*`; grid `px/py-16`, column gap 8 / row gap 16 per Figma.
- `ProductListingFilterBar.swift` — toggle **left**, `N Items` centre (`body/small`), underlined **Refine** link right (`link/medium`); height 40 (32 content + 4pt vertical padding); 32×32 toggle targets.
- `ProductListingListStyleSelector.swift` — de-duped toggle, `Style.icon(for:isSelected:)` (outline↔fill) + tests, VoiceOver `.isSelected` trait.
- `ProductListingListStyleSelectorStyleTests.swift`, `L10n.xcstrings` (+ swiftgen checksum) — results copy `%d Item(s)`.

**Shared SharedUI — deliberate DesignSystem alignment** (see cross-screen note)
- `VerticalProductCard.swift` + `VerticalProductCardConfiguration+Extension.swift` — align PLP product-cell name/brand/price typography, colours, spacing with Figma's 2-up card. Verified against Figma (medium + large card nodes): brand 12pt/#111, name 16pt/#111/lh24, price 16pt, card gap 8.
- `PriceComponentView.swift` — price weight `.semibold`(600) → `.medium`(500), matching Figma `body/medium-bold` (weight 510 → `.medium`).
- `ErrorView.swift` — inset action buttons 16pt (content margin) so CTAs aren't edge-to-edge.

### ⚠️ Cross-screen impact of the shared-component changes
`VerticalProductCard` (`.medium`) is also used by **Wishlist**; the `PriceComponentView` weight change affects **every price** — **Bag** + **PDP** too. All values checked against Figma and confirmed correct. `.small` card size is untouched, so carousels / Home are unaffected.

### Filter chips — removed for now
An earlier revision added a mock filter-chip row. Since the BFF filter-facet API doesn't exist yet, the mock introduced hardcoded user-facing strings (against the L10n rule), so it's been **removed** (along with the shared `Chip` tweaks it needed) and will be re-added with real facets in the slice that owns filtering.

### Test plan
- `./Alfie/scripts/verify.sh --skip-integration` → ✅ build + unit tests.
- Simulator (local BFF): PLP verified. Wishlist / Bag / PDP price-weight: verifying (pending a fresh build). Snapshot suite dormant repo-wide (per ticket).

[ALFMOB-427]: https://mindera.atlassian.net/browse/ALFMOB-427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ALFMOB-440]: https://mindera.atlassian.net/browse/ALFMOB-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Prathimaboggu (commented):
@khoinguyen-mindera — reviewed end-to-end against Figma [PLP `1:2568`](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=1-2568) dev mode and the ALFMOB-467 acceptance criteria.

**First, a framing note:** the description describes 4 files and says "Scoped to the PLP screen only", but the branch now carries **10 files** — the last two commits (`13916e1`, `2330360`) add a new mock-chips view and change two *shared* SharedUI components (`Chip`, `ErrorView`) that are used by other screens. None of that is in the Changes list, and the Copilot review above is stale (it reviewed `faf8d89`, and describes a `Style.iconTint(isSelected:)` that no longer exists). Please refresh the description before merge — reviewers and QA are reading a smaller PR than the one that exists.

The token migration and the filter-bar restyle are genuinely good. The chips are the problem.

---

## What I verified

Environment caveat, stated plainly: **this machine has Command Line Tools only** (`xcode-select -p` → `/Library/Developer/CommandLineTools`) — no `xcodebuild`, no iOS SDK, no `simctl`, no SwiftLint. `swift build --target ProductListing` also can't work here: AlfieKit declares only `.iOS(.v16)`, so the macOS floor defaults to 10.13 and manifest resolution fails against Firebase/Apollo/NukeUI (not a problem with this PR). **I could not re-run `verify.sh`, so your ✅ build+tests claim is unaudited by me.** Everything below is static analysis, token arithmetic, and Figma dev-mode comparison.

What I *did* verify by hand:

**Generated artifact is genuinely in sync.** I reproduced the `Run SwiftGen` build phase's checksum computation (`shasum` over the concatenation of `L10n.xcstrings` + `codegen-strings-structured.stencil` + `swiftgen.yml`, trailing newlines stripped per `$(cat …)`) and got `4c17bcf70ab7ebb5d08d04be0c7a082713e1d026` — byte-for-byte the committed value. I then sentinel-tested it (appended a marker to `L10n.xcstrings`, checksum moved to `4dd5ca41…`, restored, checksum returned) to prove the computation actually reads the file rather than trivially agreeing. `L10n+Generated.swift` correctly needs no regeneration: the key `plp.number_of_results.message` is unchanged, only its value. `LocalizationTests.testLocalizableProductListingResultsWithArgs` asserts format-specifier validity, not copy, so it still holds.

**The "value-identical" claim is true.** Checked against `SpacingProvider.swift` and `Theme+Generated.swift`:

| Swapped to | Resolves to | Old value | Identical |
|---|---|---|---|
| `theme.spacing.space0` | `spacing0` | 0 | ✅ |
| `theme.spacing.space200` (×4) | `spacing16` | 16 | ✅ |
| `theme.spacing.space400` (loader) | `spacing32` | 32 | ✅ |
| `Theme.contentContentTerciary` | `Primitives.Colours.neutrals500` | `neutrals500` | ✅ literally the same constant |

**Figma comparison — what matches exactly:**

- Results count colour: `contentContentTerciary` → `neutrals500` → `#767676`. Figma's `14 Items` is `text-[#767676]`. Exact.
- Refine colour: `linkLinkPrimaryDefault` → `neutrals800` → `#111111`. Figma `link/link-primary-default` = `#111111`. Exact.
- Typography: Figma `link/medium` = 16pt / weight 510 / lh 24 → `theme.font.link.medium(…, underline: true)`. Figma `body/small` = 12pt / 400 / lh 16 → `theme.font.body.small`. Both correct.
- Layout order: Figma is `justify-between` with toggle at x=0, count at x=155–203, link at x=294–343 — gaps of exactly 91 either side of the count, i.e. genuinely centred. Your `HStack { toggle; Spacer(); count; Spacer(); refine }` reproduces that precisely.
- **Icon mapping is exactly right**, and this is the part I'd have most expected to be wrong. Figma's leading button holds `Grid 1` (outline) and the trailing one `Grid 2 (Fill)`, with the dev annotation on the trailing button reading *"When the user clicks on this button, the PLP shows 2 items per row"* — so leading = list, trailing = grid, which is the order you used. And the raw values line up: `.listplp` = `"grid-1"`, `.grid1Fill` = `"grid-1-fill"`, `.grid` = `"grid-2"`, `.grid2Fill` = `"grid-2-fill"`. The `listplp` naming makes this look wrong at a glance; it isn't.
- Icon size: `.medium` → `Sizing.iconsIconMedium` → `spacing24` → 24. Figma icon is `size-[24px]`. Exact.
- Inter-icon gap: Figma's two 32×32 buttons sit flush (x=0, x=32) with 24pt icons centred, so the *visible* icon-to-icon gap is 4+4 = **8** — which is what `space100` gives you. The approximation you didn't flag is actually exact.
- Chips: inter-chip gap 8 = Figma `screen-size/gutter`; horizontal padding 16 = Figma `screen-size/margin`. Both exact.
- Copy: `%d Item` / `%d Items` matches Figma's `14 Items`.

**Other things I checked and found fine:**

- `.accessibilityAddTraits(isSelected ? .isSelected : [])` is correct — `AccessibilityTraits` is an `OptionSet` and `[]` is a genuine no-op, re-applied on every state change.
- The `Chip.swift` outer-padding removal **matches Figma exactly** (Figma's Chip is outer `px-0` / inner `px-16`), and I traced every call site: `Chip` had **zero** production consumers before this PR. Blast radius is the `#Preview` block only. Good change, and safer than it looks.
- `.stroke` → `.strokeBorder` is right: `RoundedRectangle` is insettable, and the previous centred stroke put half the line (a full 1pt at the 2pt selected width) outside the shape's frame.
- `Sizing.radiusRounded` = 1000 vs Figma's 56 — both clamp to a full pill at these heights. Equivalent, no action.
- `.onTapGesture` on a non-`Button` is established house style here (`RadioButton`, `Checkbox`, `ThemedSegmentedView`, `ColorSelectorComponentView`, `TabBarItemView`). Not a defect on its own — see the caveat in 🔴 2.

---

## 🔴 Blocking

### 1. Mock filter chips ship to production users

`ProductListingFilterChips.swift:44-46` — `Constants.mockFilters = ["Slim Fit", "Linen", "Cotton", …]` lives in `Sources/ProductListing/`, unconditionally compiled into the shipping binary.

Three separate house rules say this can't ship as-is:

- **`CLAUDE.md` → ❌ NEVER → "Hardcode user-facing strings"**, and `Docs/Localization.md`: *"Never hardcode user-facing strings. Always use `L10n` generated enums."* These are display copy in seven visible chips. I grepped `Sources/` — every other bare string literal in a view is inside a `#Preview` or an `AnyView(Text("Something went wrong"))` fallback. There is no precedent for this.
- **Mock data is gated behind `#if DEBUG` everywhere else.** All 19 files importing `Mocks` do so under `#if DEBUG` — including `ProductListingView.swift:4-6` in this very PR.
- **ALFMOB-467 "Out of Scope" lists "Filter/sort logic"**, and the ticket's "Scope (files)" names three files, none of which is a chips view.

And functionally: tapping a chip toggles a border and nothing else. A user in production gets seven tappable, untranslated, permanently inert filters.

**Fix:** drop `ProductListingFilterChips` from this PR and land it with the real facets in the slice that owns filtering. If you want it on screen now for design sign-off, gate the whole view behind `#if DEBUG` and move `mockFilters` into the `Mocks` module.

While it's there, two smaller things fall out of the same block: `init(filters:)` at line 14 has no call site that passes it (dead parameter — and it's the only reason `enum Constants` at line 44 can't be `private`), and `ForEach(filters, id: \.self)` at line 21 will collapse rows the moment two facets share a label, which real server-driven facets will do.

### 2. The chips are invisible to VoiceOver and to XCUITest

`ProductListingFilterChips.swift:27-28` — `.onTapGesture` with no accessibility traits.

The repo's own pattern for exactly this shape is `ThemedSegmentedView.swift:84-95`:

```swift
.accessibilityAddTraits(.isButton)
.accessibilityIdentifier(…)
.if(isSelected,
    whenTrue:  { $0.accessibilityAddTraits(.isSelected) },
    whenFalse: { $0.accessibilityRemoveTraits(.isSelected) })
.contentShape(Rectangle())
.onTapGesture { … }
```

The chips have the identifier and the tap gesture but neither trait. Concretely: VoiceOver announces "Slim Fit" as plain text with no indication it's actionable and no indication of selected state, and `ProductListingPage`-style queries (`app.buttons[AccessibilityID.…]`, per `Docs/Accessibility.md`) will never match — they'd land in `staticTexts`/`otherElements`.

This is the same gap you deliberately closed for the layout toggle in `597a5a9`; the chips landed a commit later without it.

**Fix:** add `.accessibilityAddTraits(.isButton)` plus the selected/deselected pair, following `ThemedSegmentedView`.

---

## 🟠 Should fix

### 3. Filter bar is 8pt shorter than Figma — AC 1 isn't met

`ProductListingFilterBar.swift:45` — `.frame(minHeight: theme.spacing.space400)` = **32**.

Figma's `Number of Products + View` frame (`I1:3157;3896:14984`) is **40** tall — confirmed two ways: node geometry (`x=16, y=0, w=343, h=40`) and the layout itself (`py-[spacing-xxs,4px]` around 32pt-tall buttons ⇒ 4+32+4). Tallest content in your `HStack` is a 24pt icon, so this renders at the 32 floor. The description's "bar `minHeight` **60 → space400 (32)** per Figma" is the one design claim that doesn't hold — Figma says 40.

Also worth noting the token category: 40 isn't a height in Figma, it's 32pt content plus 4pt vertical padding. Expressing it that way (see 4) reproduces it exactly and self-documents.

### 4. Toggle hit target is 24×24; Figma specifies 32×32

`ProductListingListStyleSelector.swift:14,39` — bare 24pt `ThemedIcon`s in a `Button`, spaced 8 apart.

Figma wraps each icon in a `Button/Tertiary/Small/Default/True` at `size-[32px]`, and the two sit flush (gap 0, total 64 wide). Your version is 56 wide with 24×24 hit areas. The *visible* gap is identical (see "what matches exactly"), so this is purely about the touch target — 24pt is well under both Figma's 32 and the 44pt HIG minimum. It's an improvement on the old 16pt, but it's not what the design says.

3 and 4 have one fix that resolves both:

```swift
HStack(spacing: theme.spacing.space0) {   // Figma: buttons flush, gap 0
    styleButton(…)  // each .frame(width: 32, height: 32).contentShape(Rectangle())
    styleButton(…)
}
```

with the bar as `.padding(.vertical, theme.spacing.space050)` (4) instead of `minHeight`. That lands 40pt exactly, keeps the 8pt visible icon gap, and restores the 32×32 targets.

### 5. Gap between the results row and the chips row should be 0, not 8

`ProductListingView.swift:95` — `VStack(spacing: theme.spacing.space100)` = 8.

Figma's `Refine` container is `flex-col gap-[var(--spacing/spacing-none,0px)]`, and the node geometry agrees: results row `y=0 h=40`, Chips `y=40`. Zero gap. The container instead carries `pb-[spacing-xs,8px]` *below* the chips, which the implementation doesn't have (the grid's own 16pt padding is doing something different).

### 6. `ErrorView` change silently restyles three screens outside this ticket

`ErrorView.swift:121` — `.padding(.horizontal, Primitives.Spacing.spacing16)` on the buttons `VStack`.

`ErrorView` has four call sites: `ProductListingView.swift:146`, `WebView.swift:46`, `CategoriesView.swift:93`, `ProductDetailsView.swift:531`. None of them applies its own horizontal padding, so all four now inset their full-width CTAs by 16. Three of those are outside ALFMOB-467.

The change looks *right* (16 is Figma's `screen-size/margin`), but I couldn't verify it against an `ErrorView` Figma node — flagging it as unverified rather than wrong. Please either split it into its own PR or, at minimum, add it to the description and confirm you eyeballed PDP / Shop / WebView error states.

### 7. Accessibility identifiers diverge from `Docs/Accessibility.md`

`AccessibilityID.swift:59-60`:

```swift
public static let filterChips = "productListing.filterChips"
public static func filterChip(label: String) -> String { "productListing.filterChip.\(label)" }
```

Two documented rules broken:

- **"The trailing segment is the control type"** — neighbours are `productListing.results.label`, `productListing.listStyle.grid.button`. `filterChips` has no type segment and jams two words into one camelCase segment.
- **"Identifiers are stable, unlocalised"** / **"Dynamic segments use stable model IDs"** — `filterChip(label:)` interpolates *display copy*. Today that yields `productListing.filterChip.Slim Fit` (with a space); once these become localised or server-driven facets, every identifier changes per locale and every UI test breaks.

**Fix:** `productListing.filterChips.list` and `filterChip(id:) -> "productListing.filterChip.\(id).button"` keyed on the facet's stable id.

---

## Pre-existing, not this PR (not blocking)

`ProductListingFilterBar.swift:10,32-43` — `opacity` starts at `.zero` and only animates to 1 via `.onChange(of: isLoading)`. If the bar first renders with `isLoading` already `false`, no change fires and the results count stays invisible. Untouched by this PR; worth its own ticket.

---

## Security & performance

Nothing to flag. No new network calls, no persistence, no logging, no identifiers or tokens touched. The nested horizontal-in-vertical `ScrollView` is a standard SwiftUI composition and the chip list is 7 static items, so there's no measurable cost; the `LazyVGrid` paging path is untouched.

---

## Verification gaps

1. **Build, unit tests and lint were not run by me** — no Xcode, no iOS SDK, no SwiftLint on this machine (details at the top). Your `verify.sh --skip-integration` result stands unaudited.
2. **No snapshot coverage**, as the ticket acknowledges — which means findings 3, 4 and 5 (all pure geometry) have nothing regression-guarding them. That's the accepted repo trade-off, but it does mean the Figma numbers above are the only check these values will ever get.
3. **Finding 6 is unverified against a design**: I confirmed the blast radius across four screens but not that 16pt is the intended `ErrorView` inset.

khoinguyen-mindera:
Thanks @Prathimaboggu — genuinely thorough, and thanks for hand-verifying the token arithmetic and icon mapping. Refreshed the PR description (10 files, mock chips, the two shared-component changes). Pushed `b14e0ec` addressing the rest.

**On the two "expected" items — keeping them, by design:**

- **🔴 1 — Mock chips:** intentional. The BFF filter-facet API doesn't exist yet, so the chips are a mock stand-in with local-only selection, kept on-screen so the chip design can be signed off against Figma now. They'll be replaced by real facets in the slice that owns filtering. I did fix the legitimate defects that are independent of the mock-ness (below), so it's a correct-but-mocked component rather than a broken one.
- **🟠 6 — ErrorView inset:** also intentional — it's the standard 16pt DesignSystem content margin, and applying it in the shared component is deliberate so every error screen is consistent. Documented in the description. (Blast radius is the 4 call sites you found; the change only adds side margins to already-full-width CTAs.)

**Fixed in `b14e0ec`:**

- **🔴 2 — Chip a11y:** added `.accessibilityAddTraits(.isButton)` + the `.isSelected` pair and `.contentShape(Rectangle())`, matching `ThemedSegmentedView` / the toggle in `597a5a9`.
- **🟠 3 — Filter bar height:** now `.padding(.vertical, space050)` around 32pt content = **40**, expressed as content+padding per your suggestion (was the 32 floor).
- **🟠 4 — Toggle targets:** icons now sit in **32×32** `.contentShape` frames, buttons flush (`space0`) → 8pt visible gap preserved, 32pt hit target restored.
- **🟠 5 — Results→chips gap:** VStack `space0` (0), with 8pt padding **below** the chips.
- **🟠 7 — Chip identifiers:** now index-based + `.button` suffix (`productListing.filterChip.<index>.button`) — stable, non-localised, control-type trailing segment.
- **Code smells:** dropped the dead `init(filters:)` param and switched `ForEach` to `enumerated()`/`\.offset` so duplicate labels can't collapse rows.

The pre-existing `opacity` init-state issue (results count) is a real one but untouched here — I'll raise it as its own ticket. Snapshot coverage gap on the geometry: noted, same repo trade-off.

Prathimaboggu (commented):
@khoinguyen-mindera — re-reviewed at `a9c7261` against Figma PLP [`1:2568`](https://www.figma.com/design/axx7Bz1fpQurtU6DHwVaJX/Alfie---Designs--Mobile-?node-id=1-2568) dev mode and the ALFMOB-467 acceptance criteria. Scoping this to the three things I think need action before merge.

**Environment, stated plainly:** Command Line Tools only — `xcode-select -p` → `/Library/Developer/CommandLineTools`, `xcodebuild -version` → *"requires Xcode"*, no `simctl`. There's also no SwiftLint binary in the repo (`Tools/` contains only `DesignTokenGen`), so the `SwiftLint` build phase warns-and-skips on every local build — not just mine. **I did not build, test or lint.** Everything below is static analysis, token arithmetic and Figma dev-mode comparison; your `verify.sh --skip-integration` result stands unaudited by me.

**Process note:** the description says 10 files; the branch carries **13**. `a9c7261` adds `PriceComponentView.swift`, `VerticalProductCard.swift` and `VerticalProductCardConfiguration+Extension.swift`, and it appears in neither the Changes list nor your reply above — so it has had no review from anyone. That commit is most of what follows.

---

## Status of the seven earlier points

| # | Status |
|---|---|
| 🔴 1 mock chips in production | Declined by you — noted, not re-litigating |
| 🔴 2 chip a11y traits | **Fixed** |
| 🟠 3 filter bar height 40 | **Fixed — verified exact** |
| 🟠 4 32×32 toggle targets | **Fixed — verified exact** |
| 🟠 5 results→chips gap 0 | Fixed, but the `pb` that came with it overshoots → **S1** |
| 🟠 6 ErrorView blast radius | Declined — holding a correction for follow-up |
| 🟠 7 chip identifiers | Partially fixed; the new form breaks a different documented rule → **S2** |

Credit where it's due on the two that had numbers attached, both exact:

- **Bar height (🟠 3).** `ProductListingFilterBar.swift:45` — `.padding(.vertical, theme.spacing.space050)`; `space050` → `spacing4` → 4. Tallest content is the 32pt toggle frame ⇒ 4+32+4 = **40**. Figma `I1:3157;3896:14984` is `py-[var(--spacing/spacing-xxs,4px)]` around `size-[32px]` buttons ⇒ 40. Expressed as content+padding as suggested.
- **Toggle targets (🟠 4).** `ProductListingListStyleSelector.swift:44` — `.frame(width: theme.spacing.space400, height: theme.spacing.space400)` (32×32) + `.contentShape(Rectangle())`, buttons flush via `space0` (line 14). Figma: two `Button/Tertiary/Small/Default/True` at `size-[32px]` inside `View Selection` (`flex items-center`, no gap class ⇒ 0), 24pt icons centred ⇒ the 8pt visible icon gap is preserved.

And separately, from `a9c7261`: **the `.semibold` → `.medium` price weight is the correct Figma reading**, and it's the one I'd have most expected to be wrong. Figma's price token is `body/medium-bold` at **weight 510** (`SF_Pro:Medium`); the repo's own `UIFont.Weight(tokenWeight:)` maps 500 → `.medium`, and there is no `bodyMediumBold` token to reach for. `.semibold` (600) was the wrong end.

---

## 🔴 B1 — `a9c7261` restyles shared components the ticket puts explicitly out of scope; Wishlist, Bag and PDP change unverified

ALFMOB-467 → **Out of Scope**: *"`VerticalProductCard` internals — it is a shared SharedUI component with its own rollout; this slice only configures its size (`.medium`/`.large`)."*

`a9c7261` changes the internals. Diffed against `main`:

| Property | Card size | `main` | head | Screens affected |
|---|---|---|---|---|
| `verticalInterSpacing` | `.medium` | 12 | **8** | PLP grid, **Wishlist** |
| | `.large` | 16 | **8** | PLP list |
| name font | `.medium` | `body.small` 12pt | **`body.medium` 16pt** | PLP grid, **Wishlist** |
| name colour | `.medium` / `.large` | `neutrals500` #767676 | **`neutrals800` #111111** | PLP grid + list, **Wishlist** |
| name reserved height | `.medium` | 2 × ≈14.3 ≈ **28.6** | 2 × 24 = **48** | PLP grid, **Wishlist** |
| designer font | `.large` | `body.medium` 16pt | **`body.small` 12pt** | PLP list |
| price weight | all | `.semibold` (600) | **`.medium` (500)** | PLP, **Wishlist, Bag, PDP** |

- `WishlistViewModel.swift:55` configures `size: .medium, hideDetails: false` — so every row of the Wishlist grid gets a larger, darker, ~19pt-taller name block plus tighter internal spacing.
- `PriceComponentView.swift:207-212` is the only weight source for `.default` and sale-final prices, and it's consumed by `VerticalProductCard` (PLP, Wishlist, carousels), `HorizontalProductCard` (Bag — `BagView.swift:21`) and `ProductDetailsView.swift:362`. **Every price in the app** changes weight.

To be clear about what I'm *not* saying: I checked each value against Figma's 2-up card (`I725:12955;3003:11335`) and they're **right** — brand `label/small` 12/400/#111, name `body/medium` 16/400/lh 24/#111, price 16/#111, card `gap-[var(--spacing/spacing-xs,8px)]` = 8. And `.small` is provably untouched (`nameFont`/`nameColor` keep `body.small`/`neutrals500`, `nameLineHeight` returns the font's own lineHeight so the `lineSpacing` expression computes to 0, `verticalInterSpacing` was already 8), so `ProductCarouselView.swift:51` and the home carousels are unaffected — that's what keeps this to three screens instead of five.

The problem is scope and verification: three screens outside ALFMOB-467 change appearance, the description doesn't mention them, and the test plan covers only PLP.

**Fix:** split `a9c7261` into its own PR against the `VerticalProductCard` / price rollout ticket — or, if it stays, add Wishlist + Bag + PDP to the description and confirm you checked all three in the simulator.

---

## 🟠 S1 — chips → first card row is now 24pt; Figma says 8 (the fix for 🟠 5 overshot)

`ProductListingView.swift:105` adds `.padding(.bottom, theme.spacing.space100)` (8) below the bar + chips VStack. `ProductListingView.swift:86` still applies `.padding(theme.spacing.space200)` (16, **all edges**) to the grid.

Figma:
- `Refine` (`1:3157`): `pt-[var(--spacing/spacing-none,0px)] pb-[var(--spacing/spacing-xs,8px)]`
- `Vertical Product List` (`725:12946`): `px-[var(--screen-size/margin,16px)] **py-[var(--spacing/spacing-none,0px)]**`
- Node geometry agrees: `Refine` is `y=95 h=88` (ends at 183); `Vertical Product List` starts at `y=183`.

So Figma's chips→card gap is exactly the 8 carried on the Refine block, *because the list contributes 0*. The implementation adds the 8 **and** keeps the grid's 16pt top ⇒ **24**. It was 16 before this PR, so the fix moved it further from Figma rather than closer — the grid's own top padding needs to come off at the same time.

```swift
// ProductListingView.swift:86 — Figma: px-16, py-0
.padding([.horizontal, .bottom], theme.spacing.space200)
```

---

## 🟠 S2 — chip identifiers now use array indexes, the one form `Docs/Accessibility.md` explicitly forbids

`AccessibilityID.swift:60`:
```swift
public static func filterChip(index: Int) -> String { "productListing.filterChip.\(index).button" }
```

`Docs/Accessibility.md:26` — *"Dynamic segments use stable model IDs — **never array indexes**."*
And in Gotchas — *"**Dynamic identifiers must use stable model IDs.** `forEach { i, item in id_\(i) }` breaks when ordering changes; use `\(item.id)` instead."*

The house pattern is `ThemedSegmentedView.swift:85`: `"\(AccessibilityId.segmentedOption)-\(segment.id)"`.

The `.button` suffix and the un-localised requirement are both satisfied now, so this is real progress on my earlier point — but the index is precisely the anti-pattern the doc names, and facet *ordering* is the thing most likely to change the moment the BFF drives these.

**Fix:** give the mock facets stable ids and key on those, so the identifier shape survives the swap to real facets unchanged:

```swift
private let filters = [(id: "slim-fit", label: "Slim Fit"), (id: "linen", label: "Linen"), …]
// AccessibilityID
public static func filterChip(id: String) -> String { "productListing.filterChip.\(id).button" }
```

Also still open from 🟠 7: `AccessibilityID.swift:59` — `filterChips = "productListing.filterChips"` has no control-type trailing segment (neighbours are `productListing.results.label`, `productListing.listStyle.grid.button`). `productListing.filterChips.list` fixes it.

---

For the record so they're known rather than missed: I found a few further **non-blocking** Figma deltas I'm holding back for a follow-up rather than loading into this pass — `Chip`'s own label typography and border colour (12pt/#4A4A4A/#E9E9E9 vs Figma's 16pt/#111/#A1A1A1), the grid column gap being 8 in Figma rather than 16, and a correction to the ErrorView blast radius (3 call sites, and PLP isn't one of them since `buttons` defaults to `[]`). None of those blocks; happy to raise them as their own tickets or a second review pass, whichever you prefer.

khoinguyen-mindera:
Thanks again @Prathimaboggu — and for verifying #3/#4 to the pixel. Pushed `41e2c65`.

**Fixed:**
- **🟠 S1 — chips→card gap:** dropped the grid's top padding — `ProductListingView.swift` now `.padding([.horizontal, .bottom], space200)` (Figma `px-16 / py-0`), so the only gap is the 8 carried on the Refine block. Net 8, not 24.
- **🟠 S2 — chip identifiers:** replaced the array index with stable facet ids — `filters` is now `[(id:, label:)]` and `filterChip(id:) -> "productListing.filterChip.\(id).button"`, so the shape survives reordering and the swap to real facets. Container is now `productListing.filterChips.list` (control-type trailing segment).
- **Non-blocking grid gap:** `GridItem` column spacing 16 → **8** to match Figma's `16px 8px` (row 16 / column 8). Row spacing stays 16.

**🔴 B1 — the card commit (`a9c7261`), keeping it, now documented:** you're right that it touches shared internals the ticket scopes out, and thanks for confirming every value against the 2-up card. It's a deliberate DesignSystem alignment and the price-weight read is intentional (`body/medium-bold` = 510 → `.medium` via `UIFont.Weight(tokenWeight:)`, as you traced). I've updated the description to call out the cross-screen impact explicitly — **Wishlist** (card typography/spacing), and **Bag + PDP** (price weight) — and I'm doing a simulator pass over those three before merge. If you'd still prefer it split out to the `VerticalProductCard`/price rollout ticket, I'm happy to — your call.

**Non-blocking deltas you're holding** (Chip label typography/border colour, and the ErrorView blast-radius correction — agreed, 3 sites and PLP isn't one since `buttons` defaults to `[]`): I'll raise these as their own follow-up tickets rather than widen this PR further. The pre-existing `opacity` results-count issue too.

khoinguyen-mindera:
Update: **removed the mock filter chips** (`30c533c`).

Rather than keep a mock that hardcodes user-facing strings against the L10n rule (your 🔴 1), we've pulled the chip row entirely — it'll come back with the real BFF facets in the slice that owns filtering. That also **fully reverts the shared `Chip.swift`** change (back to `main`, so your out-of-scope flag on it is gone) and drops the chip accessibility identifiers, so 🔴 1 / 🟠 2 (chip a11y) / 🟠 7 (chip ids) are moot now.

Net effect on this PR: down to **10 files**, scoped to the PLP token/restyle work plus the shared card/price DesignSystem alignment (B1) — which I've verified against Figma's medium + large card nodes (brand 12/#111, name 16/#111/lh24, price 16/`.medium`, gap 8) and documented the Wishlist/Bag/PDP impact in the description. Still happy to split the card/price commit out if you'd prefer.

Prathimaboggu (commented):
@khoinguyen-mindera — third pass, scoped to the two commits since my last review at `a9c7261` (head is now `30c533c`). This one is better corroborated than the previous two: GitHub Actions run [30878581569](https://github.com/Mindera/Alfie-iOS/actions/runs/30878581569) built and ran the unit suite on the exact head commit and passed (`Test Succeeded`, 0 failures — `ProductListingTests` 23, `SharedUITests` 72, `ProductDetailsTests` 76, `WishlistTests` 3), so the `verify.sh` result no longer stands on its own. I still did not build locally (CLT-only machine, no iOS SDK, no simulator), and there is no lint step in `.github/workflows/` — SwiftLint runs only as an Xcode build phase. Everything below is static analysis plus Figma dev-mode node geometry.

## Status of the three points from `a9c7261`

| # | Status |
|---|---|
| 🔴 B1 — shared card/price restyle out of scope | Declined, now documented in the description |
| 🟠 S1 — chips→card gap should be 8 | **Fixed in `41e2c65`, then regressed in `30c533c`** — below |
| 🟠 S2 — index-based chip identifiers | **Moot** — chips and both identifiers deleted |

---

## 🟠 The S1 fix was undone by the chips-removal commit; bar→grid gap is now 16, Figma says 8

`ProductListingView.swift:86` — back to `.padding(theme.spacing.space200)` (all edges, so 16pt top).

`41e2c65` fixed this correctly: the grid became `.padding([.horizontal, .bottom], space200)` (0 top), with the 8 carried on the bar block's `.padding(.bottom, space100)`. Net 8, exactly Figma. `30c533c` then reverted line 86 to all-edge padding *and* deleted the `.padding(.bottom, space100)` along with the VStack that held the chips — so the 8 became **16**. The commit message covers the chips and the `Chip` revert; the padding revert is an undocumented side effect of it.

Figma, from node geometry rather than class names:

- `Refine` (`1:3157`): `h=88`; children `Number of Products + View` at `y=0 h=40` and `Chips` at `y=40 h=40`, so content ends at 80 ⇒ **8pt bottom padding on the container**.
- `Vertical Product List` (`1:3158`): starts at `y=183`, exactly where `Refine` ends; its inner `Frame 4656` sits at `y=0` ⇒ **`py-0`**, contributing nothing to the gap.

The 8 is the `Refine` *container's* padding, not something the chips carried — removing the chips doesn't remove it. So with the chips gone, the bar-bottom→first-card gap should still be 8.

This also makes one claim in the description wrong: **"grid `px/py-16` … per Figma"**. Figma is `px-16 py-0`.

```swift
// ProductListingView.swift:86 — Figma: px-16, py-0
.padding([.horizontal, .bottom], theme.spacing.space200)
```

```swift
// ProductListingView.swift:95 — Figma: Refine container pb-8
ProductListingFilterBar(…) { … }
    .padding(.bottom, theme.spacing.space100)
```

---

Nothing blocking. That gap is the only defect in the two new commits.

## Verification gap

The B1 blast radius is still unconfirmed. The test plan reads "Wishlist / Bag / PDP price-weight: **verifying** (pending a fresh build)" — and CI green does not cover appearance. Three screens outside ALFMOB-467 change (Wishlist card typography and spacing; Bag and PDP price weight), with no snapshot coverage behind them. That's the item I'd gate merge on.

Prathimaboggu (approved):

PR #103, opened 2026-07-30
ALFMOB-468: Modern Design Rollout — PLP Refine & Sort sheet

khoinguyen-mindera:
## What & why
Slice 2/3 of [ALFMOB-440](https://mindera.atlassian.net/browse/ALFMOB-440) (Shop rollout). Restyles the **PLP Refine & Sort bottom sheet** to the modern Figma design — visual-only, no behavioural change. Follows the pattern shipped in ALFMOB-437/438/439.

**Stacked on** `ALFMOB-467-plp-modern-design` (slice 1) — review/merge that first.

Ticket: https://mindera.atlassian.net/browse/ALFMOB-468

## Changes
- `ProductListingFilter.swift` — the sheet: header, dividers, list-style row, apply button spacing.
- `SharedUI/.../SortByView.swift` — the sort pill list rendered in the sheet (only production consumer is this sheet; also a debug demo).

Token migration:
- `Primitives.Spacing.spacing*` → `theme.spacing.space*` (8→100, 12→150, 16→200, 24→300)
- `Primitives.Colours.neutrals800/900` → `Theme.contentContentPrimary`; unselected pill border `neutrals100` → `Theme.borderSoft`
- Dropped hardcoded font sizes (`listStyleFontSize=18`, `titleFontSize=18`) → theme font ramp (`theme.font.body.medium`)

## Intentional visual deltas (per AC)
- Section labels "List style"/"Sort By": 18pt → 16pt (`body.medium` natural size).
- Unselected sort-pill border: neutrals100 → neutrals200 (nearest semantic border token).

## Out of scope
PLP grid / filter bar (separate slice), filter/sort logic, dark mode.

## Verification
- `./Alfie/scripts/verify.sh --skip-integration` → ✅ build + unit tests pass.
- No behavioural change; `SortByHelper.swift` untouched (pure data). Snapshots dormant repo-wide (consistent with 437–439).

## Acceptance criteria
- [x] Sheet consumes `Theme.*` / `theme.spacing.*` / SharedUI components — no bespoke hardcoded colours/spacing/font sizes.
- [x] No regression in applying/clearing filters or changing sort.
- [x] `verify.sh` passes.

[ALFMOB-440]: https://mindera.atlassian.net/browse/ALFMOB-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alfie/AlfieKit/Sources/SharedUI/Components/Slider/RangeSliderStyle.swift
```diff
@@ -0,0 +1,154 @@
...
+        return lowerDistance < upperDistance ? .lower : .upper
+    }
+
+    // MARK: - Text field round-trip
+
+    /// Whole-unit text for a field. A nil value renders empty — never `0`, which would read as a
+    /// real bound (ALFMOB-481).
+    func fieldText(for value: Double?) -> String {
+        guard let value else { return "" }
+        return String(Int(value.rounded()))
```
khoinguyen-mindera: **🔴 Critical — `Int(Double)` traps here; the price field can reach it.**

Nothing in the chain limits what the user types. `fieldValue(from:)` (line 124 below) does `Double(digits)` with no ceiling, writes it straight to `pendingMinPrice`, and the very next render lands on this line.

```
19 digits → 1e19        Int.max as Double is 9.223372036854776e18
Fatal error: Double value cannot be converted to Int because the result would be greater than Int.max
```

I compiled and ran this — it aborts with SIGILL (exit 133). 400 digits gives `inf`, which traps identically. Reachable by holding a key on the `.numberPad` field.

**Fix:** clamp at the parse boundary in `fieldValue(from:)` — a price filter has no business accepting a value orders of magnitude above `bounds.upperBound` — and use `Int(exactly:)` with a fallback here rather than the trapping initialiser.

A reproduction test belongs in `RangeSliderStyleTests` (see my comment there).

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefineViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+    /// The glyph the price fields show as their affix. Derived from the BFF's currency code, not
+    /// the `$` the Figma panel draws — the PLP it filters is priced in the store's own currency.
+    var currencySymbol: String? {
+        priceBounds.map { CurrencyFormatter.symbol(for: $0.currencyCode) }
+    }
+
+    /// Trailing summary on the Price row; nil when the dimension is unfiltered.
+    var priceSummary: String? {
+        guard let symbol = currencySymbol, hasActiveFilters else { return nil }
+        let minimum = pendingMinPrice.map { "\(symbol)\(Int($0))" }
```
khoinguyen-mindera: **🔴 Critical — same `Int(Double)` trap as `RangeSliderStyle.fieldText`.**

`Int($0)` traps for any `pendingMinPrice`/`pendingMaxPrice` above `Int.max`, which the unbounded price field can produce. Same crash, second site.

**Fix:** use `CurrencyFormatter.string(amount: Decimal($0), currencyCode:)` instead of `"\(symbol)\(Int($0))"`. That kills the trap *and* fixes symbol placement — string interpolation hardcodes symbol-before-amount, which is wrong for locales that suffix the glyph, and `CurrencyFormatter` is already the project's single source of truth for price strings. It is also what the slider itself announces to VoiceOver via `valueDescription`, so the row summary and the slider would stop disagreeing (`£40` vs `£40.00`).

Note `Int()` truncates while `fieldText` rounds, so today these two can differ by 1 for any non-integer value.

Alfie/AlfieKit/Sources/ProductListing/UI/ProductListingFilter.swift
```diff
@@ -1,99 +1,261 @@
...
-    @State private var sort: SortByType
+    /// Scoped to this sheet rather than added to `ProductListingRoute`: the refine surface is
+    /// presented *over* the PLP stack, so its sub-screens are not flow-level navigation and must
+    /// not share the PLP's push stack (ALFMOB-476).
+    private enum RefineRoute: Hashable {
+        case price
+        case sort
+    }
+
+    @StateObject private var viewModel: RefineViewModel
```
khoinguyen-mindera, line 16: **🟠 Important — the sheet snapshots `priceBounds` at presentation; a late fetch never appears.**

`@StateObject`'s `wrappedValue` autoclosure runs **once per presentation**, capturing the `let priceBounds` passed into `RefineViewModel`. But:

- the Refine button is live while `state.isLoadingFirstPage` (`ProductListingFilterBar` renders with `isLoading:`, it isn't disabled), and
- `loadPriceBoundsIfNeeded()` races the product query — both are fired from `viewDidAppear`.

Tap Refine before the bounds land and `priceBounds == nil`, so `rows` renders **no Price row** — and it cannot recover, because `didRequestPriceBounds` correctly prevents a re-query. The data arrives; this sheet instance just can't see it. The user has to dismiss and reopen.

Usually the small bounds query wins the race, so this will be intermittent — which is the worst kind of bug to leave in.

**Fix:** observe the parent view model from the sheet, or re-seed on `.onChange(of: viewModel.priceBounds)`.

No test covers a late-arriving bounds value — `ProductListingViewModelTests` tests the fetch, but nothing tests the sheet's seeding against it.

Alfie/AlfieKit/Sources/ProductListing/UI/ProductListingFilterChips.swift
```diff
@@ -0,0 +1,44 @@
...
+
+// MARK: - ProductListingFilterChips
+
+/// Horizontal row of filter chips shown under the PLP filter bar. The labels are mock stand-ins for
+/// the (not-yet-available) server-driven filter facets; selection is local-only, with no filtering
+/// behaviour. Replace `mockFilters` + the local selection with the real facets once the BFF exposes them.
+struct ProductListingFilterChips: View {
+    @State private var selectedIndices: Set<Int> = []
+
+    private let filters = ["Slim Fit", "Linen", "Cotton", "Straight Fit", "Wool", "Regular Fit", "Silk"]
```
khoinguyen-mindera: **🟠 Important — seven hardcoded English user-facing strings shipped to production.**

Direct violation of the project's NEVER rule (`CLAUDE.md`: *"Hardcode user-facing strings"*). These aren't behind a debug flag or a feature toggle — this row renders under the filter bar on **every PLP**, is tappable, shows selection state, and filters nothing.

I know the plan flags this (ALFMOB-480, *"stays mocked and filters nothing — flag to QA"*), but that is an argument for gating it, not for shipping untranslatable controls that look functional to a user. A chip that highlights when tapped and changes no results reads as a bug, not as a placeholder.

**Fix:** put the row behind a feature flag (Firebase Remote Config is already wired) until the BFF exposes real facets, or move the labels into `L10n.xcstrings` if it genuinely ships as-is.

Alfie/AlfieKit/Sources/SharedUI/Components/Slider/RangeSlider.swift
```diff
@@ -0,0 +1,273 @@
...
+    private func field(for thumb: RangeSliderStyle.Thumb, inputs: RangeSliderConfiguration.Inputs) -> some View {
+        TextInput(
+            configuration: .init(
+                text: textBinding(for: thumb),
+                label: label(for: thumb),
+                // The category bound is the placeholder; an empty field means that side is
+                // unfiltered, so it must never pre-fill with a real-looking value.
+                placeholder: style.fieldText(for: bound(for: thumb)),
+                prefix: inputs.prefix,
+                keyboardType: .numberPad,
```
khoinguyen-mindera, line 178: **🟠 Important — `.numberPad` has no return key, and nothing here dismisses the keyboard.**

The Price sub-screen adds no `ToolbarItemGroup(placement: .keyboard)`, no tap-to-dismiss, and no `.scrollDismissesKeyboard`. Once a field is focused the keyboard covers the slider the user is trying to set, and there is no Done affordance to get it back.

Not a hard trap — the pinned CTA rides above the keyboard and the back chevron stays reachable — but the primary control of the screen is occluded with no obvious way out.

**Fix:** the repo already has `View.hideKeyboard()` in `Utils/Extensions/View+Extension.swift:55`. Wire it to a keyboard-toolbar Done item, or to a background tap on `RefinePriceView`.

Flagging at medium confidence — this is a claim about runtime behaviour that I reached by reading the code, not by exercising the sheet. If the simulator pass covered focusing a price field and dismissing the keyboard, say so and I'll withdraw it.

Alfie/AlfieKit/Sources/BFFGraph/CodeGen/Schema/schema.graphqls
```diff
@@ -123,6 +128,8 @@ type MoneyRange {
 type Mutation {
   addToCart(input: AddToCartInput!): Cart!
   createCart(input: CreateCartInput!): Cart!
+  removeFromCart(cartId: String!, lineId: String!): Cart!
+  updateCart(input: UpdateCartInput!): Cart!
```
khoinguyen-mindera, line 132: **🟠 Important — unrelated cart schema pulled in by the sync commit.**

`79b3ddb` ("sync BFF GraphQL schema") also brought in `removeFromCart`, `updateCart`, and further down `UpdateCartInput` / `UpdateCartLineInput`. Nothing in this PR references any of them, and they land in the **bag** domain — one of this project's high-rigor areas.

Blind schema syncs have bitten this repo before (the BFF auth guard). Please trim the sync to `categoryPriceRange` + `CategoryPriceRange`, or land the schema bump as its own PR where the cart surface gets reviewed on purpose.

Alfie/AlfieKit/Sources/ProductListing/Helpers/SortByHelper.swift
```diff
@@ -5,6 +5,5 @@ enum SortByHelper {
         SortByItem(value: .priceDesc, title: L10n.SortBy.PriceHighToLow.title, icon: .chartDownTrend),
         SortByItem(value: .priceAsc, title: L10n.SortBy.PriceLowToHigh.title, icon: .chartUpTrend),
         SortByItem(value: .alphaAsc, title: L10n.SortBy.AlphaAsc.title, icon: .aCircle),
```
khoinguyen-mindera, line 7: **🔵 Nit — orphaned localisation key after the `.alphaDesc` removal.**

Dropping `Z_A` leaves `L10n.SortBy.AlphaDesc.title` and the `sort_by.alpha_desc.title` entry in `L10n.xcstrings` with no remaining reference.

The plan explicitly says to keep the `z-circle` asset, but says nothing about the string — this looks like an oversight rather than a decision. Per the repo's own rule about cleaning up what your change orphans, worth removing in this PR.

Alfie/AlfieKit/Tests/SharedUITests/RangeSliderStyleTests.swift
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    func test_a_value_renders_as_a_whole_number() {
+        XCTAssertEqual(sut.fieldText(for: 40), "40")
+        XCTAssertEqual(sut.fieldText(for: 39.6), "40")
+    }
+
+    func test_clearing_a_field_clears_the_bound() {
+        XCTAssertNil(sut.fieldValue(from: "", for: .lower))
+        XCTAssertNil(sut.fieldValue(from: "abc", for: .upper))
```
khoinguyen-mindera, line 145: **🔵 Nit — this is the coverage gap that let the crash through.**

The round-trip tests cover `""` and `"abc"`, but never a large-magnitude input — which is exactly the case that traps in `fieldText`. The suite is otherwise genuinely thorough (the floor/ceil boundary test and the collapsed-range tie-break tests are the right instincts), which makes the omission stand out.

Add the reproduction case alongside the fix:

```swift
func test_a_value_too_large_for_Int_does_not_crash_the_field() {
    let huge = sut.fieldValue(from: String(repeating: "9", count: 19), for: .lower)
    XCTAssertNotNil(huge)
    XCTAssertFalse(sut.fieldText(for: huge).isEmpty)  // traps today
}
```

khoinguyen-mindera:
### Correction to my review above

I reviewed against a stale snapshot of the PR description. Two claims in my review body are wrong and I retract them:

1. **"Spec drift / the PR description describes a different change."** Withdrawn entirely. The current description is thorough and accurate — it covers the Refine architecture, the new SharedUI components, the data layer, the money-units constraint, the deliberate Figma deviations, and it explicitly explains why this landed inside 468 and why the two ALFMOB-467 commits ride along. My point about `_status.md` being stale is already covered by an existing review comment, so nothing of mine survives here.

2. **"The sheet has never been exercised in the simulator."** Withdrawn. That came from the plan doc's Outcome section; the description reports `verify.sh` → FULL VERIFICATION PASSED including 10 integration tests against a live BFF, plus a device pass. My blocking item #3 ("that is not a coincidence") was built on the stale premise — drop it. I have edited the `RangeSlider.swift` comment accordingly.

Also worth correcting: my review body said I had not run `verify.sh` locally and that CI ran unit tests only. That is still true of *my* verification, but it should not be read as doubt about yours.

**What still stands, unchanged:**

- 🔴 `Int(Double)` trap in `RangeSliderStyle.fieldText` and `RefineViewModel.priceSummary` — reachable from the numeric keypad, verified with a compiled repro. This is the one blocking item.
- 🟠 `priceBounds` snapshotted at sheet presentation (`ProductListingFilter.swift:16`).
- 🟠 Unrelated cart schema in `schema.graphqls`.
- 🟠 `.numberPad` keyboard dismissal — now flagged explicitly as unverified-by-me rather than leaning on the plan doc.
- 🔵 Orphaned `sort_by.alpha_desc.title`; missing large-input test case.

On the mocked chip row (`ProductListingFilterChips.swift`): the description discloses it clearly and asks that it not be logged as a bug, which I had not seen. I would still raise it — disclosure in a PR description doesn't reach the user looking at seven tappable English chips that filter nothing, and the hardcoded strings are a `CLAUDE.md` rule either way — but treat it as a question for you rather than a finding you failed to consider.

And to say the part that survives all of this: the money-unit handling is the best part of the diff. `Decimal` throughout, one conversion point, floor-min/ceil-max applied consistently across bounds *and* stepping, and tests that would actually catch a fixed ÷100. That is the thing most likely to be got wrong here, and it is right.

Alfie/AlfieKit/Sources/SharedUI/Components/Slider/RangeSlider.swift
khoinguyen-mindera: Confirmed — and I cannot withdraw it for you: the simulator pass did **not** cover this. I injected text through the automation tool rather than tapping into a focused field, so no keyboard ever appeared in any of my screenshots. Your read of the code was right.

Fixed in `f848e35`: `RefinePriceView` now carries a `ToolbarItemGroup(placement: .keyboard)` with a Done button plus tap-to-dismiss on the background, both using the existing `View.hideKeyboard()`. Still verified by reading rather than by running, so worth a look during your re-review.

Alfie/AlfieKit/Sources/BFFGraph/CodeGen/Schema/schema.graphqls
khoinguyen-mindera: Leaving the cart schema in, with the reasoning stated rather than silently.

The generated types are inert until an operation references them — nothing in this PR or on main queries `removeFromCart` / `updateCart`, so no cart behaviour changes here. Trimming `79b3ddb` would mean hand-editing a generated schema file, which reintroduces exactly the drift the sync exists to prevent, and re-running the sync would pull the same surface back.

Your underlying point stands though: the bag team should review that surface on purpose, not inherit it. Flagged here so it is on the record when they land an operation against it.

khoinguyen-mindera:
## Review round 1 addressed — `7599841`

Every finding acted on. `./Alfie/scripts/verify.sh` → **FULL VERIFICATION PASSED** (build + unit + 10 integration tests against a live BFF), 18 new tests, 0 failures.

### 🔴 The crash — fixed, `9649155`
Reproduced your abort independently before touching it (exit 133, SIGILL). Both sites fixed:
- `fieldValue(from:)` now saturates typed input at `RangeSliderStyle.maximumFieldValue` (999,999,999) and pins `.infinity` there rather than returning nil, so holding a key caps the field instead of clearing it.
- `fieldText` uses `Int(exactly:)` with an empty-string fallback — defence in depth for a value arriving from anywhere but the parser.
- `priceSummary` now goes through `CurrencyFormatter.string(amount:currencyCode:)` as you suggested. That kills the trap, fixes symbol placement for suffixing locales, and resolves the truncate-vs-round disagreement you spotted between `Int()` and `fieldText`.

Your suggested reproduction test is in, plus infinity, saturation, round-trip and non-finite cases.

### 🟠 Important — all fixed
| Finding | Fix |
|---|---|
| `priceBounds` snapshotted at presentation | `RefineViewModel.priceBounds` is now `@Published`, re-seeded via `.onChange`. Test covers late arrival. `7a05429` |
| Stale in-flight page overwrites filtered result (Copilot) | Added a result-set generation counter; `refresh`, `loadMore` and first-page all refuse to commit across a filter change. Test suspends a page mid-flight and asserts the stale one loses. `7a05429` |
| Hardcoded chip strings | Seven `plp.quick_filter.*` keys. `f0e12db` |
| `.numberPad` dismissal | Keyboard toolbar Done + tap-to-dismiss — see my reply on that thread, your premise was right. `f848e35` |
| Cart schema | Left in, reasoning on that thread. |

### 🔵 Nits + Copilot's suppressed set
Orphaned `sort_by.alpha_desc.title` removed; large-input test added; duplicate "Sort By" heading gone (`SortByView.title` is now optional); `filterInput` preserves untouched dimensions and Remove All clears them too; three VoiceOver gaps fixed (unbounded sides announce "No minimum"/"No maximum", the currency prefix is folded into the field label, and `TextInput` no longer overrides with an empty label); `CategoryPriceRangeConverterTests` covers the GraphQL→domain bridge including the min/max transposition case.

Stale plan docs refreshed — `_status.md`, `scope.md` and the plan's verification section.

### On splitting the PR
Not doing it now — the restyle commit is already the base of the branch and separating it would mean rewriting 17 commits and re-basing an open PR. But your diagnosis is right: bundling a token migration with a money path is how the crash survived. Worth a note in the rollout convention so later ALFMOB-427 slices don't repeat it.

Alfie/AlfieKit/Sources/BFFGraph/CodeGen/Schema/schema.graphqls
khoinguyen-mindera: Following up with the detail, and correcting myself: I called this "a generated schema file" above. It is not. `sync-bff-schema.sh:60` is a plain `cp` of the BFF's own `schema.graphqls` into `CodeGen/Schema/` — it is an **input** to Apollo codegen, not an output. The generated artifacts are `BFFGraph/API/` and `BFFGraph/Mocks/`. It is also not in the project's `generated_paths`, so the never-edit-generated rule does not forbid touching it. Trimming it is permitted; it is just wrong for a different reason than I gave.

**Why not trim.** The file is a mirror of the server contract. Deleting the cart lines makes it assert the BFF has no `updateCart` when it does, and the next sync silently restores them. That buys one tidy PR at the cost of a file that lies about the server.

**Why not split into its own PR.** Two reasons:

1. **It would have to be stacked.** `CategoryPriceRangeQuery.graphql` cannot codegen without `categoryPriceRange` in the schema, so #103 could not merge until the schema PR did. This branch was already stacked on ALFMOB-467 and that ordering caused enough confusion to be worth not repeating.
2. **It presents the same diff.** A standalone "sync BFF schema" PR shows a reviewer the identical 24 lines, just in a different tab. Splitting relocates the review rather than creating it.

**What the surface actually is.** 24 added lines: 5 for `categoryPriceRange`, 19 for cart. I checked what those 19 produced — **no generated Swift at all**. Apollo only emits types an operation references, and nothing references them:

```
$ ls BFFGraph/API/Schema/InputObjects/ | grep -i cart   -> no cart input objects generated
$ grep -rl "removeFromCart|updateCart|UpdateCartInput" BFFGraph/API BFFGraph/Mocks
                                                       -> no generated Swift references cart
```

So the cart surface exists purely as text in the schema file. No types, no call sites, no behaviour, nothing for this PR to get wrong.

**Where your point survives all that**, and it does: the risk was never this PR, it is that whoever writes the first cart operation builds on a surface nobody looked at on purpose — in a high-rigor domain. Splitting the PR would not have fixed that either; a reviewer skimming a schema-sync PR is exactly the blind review you are warning about. What fixes it is the bag team reviewing the contract deliberately before they build on it. This thread is the record that it arrived unreviewed, so that is inherited knowingly rather than silently.

Happy to raise a ticket against the bag domain to review the cart schema surface if you would rather it not live only in a PR thread.

khoinguyen-mindera: Agreed on all of it — resolving this, and downgrading my own severity: I filed it as **Important** and it does not earn that. It should have been a Nit. "Unreviewed text in a contract mirror" is not the same class of thing as a state-transition hole, and I mis-ranked it.

I independently checked both of your claims before replying and they hold:

- `git diff origin/main...HEAD -- schema.graphqls | grep '^-'` → **empty**. The sync was purely additive: no removals, no nullability changes.
- No generated Swift references `removeFromCart` / `updateCart` / `UpdateCartInput` anywhere in `BFFGraph/API` or `BFFGraph/Mocks`.

Your point about trimming is the one I had wrong. I was treating the file as a changelog when it is a mirror — deleting the cart lines would make it assert the BFF has no `updateCart` when it does, and the next sync restores them. A file that lies about the server is worse than 19 unread lines.

**On the wider question of leaving the auto-sync as-is** — I went looking for the silent-drift vector and it is narrower than I expected:

- **Removed field, or `T!` → `T` on a field an operation selects** → regenerated Swift changes shape, converter fails to compile. **Loud.** This is the majority of breaking changes.
- **New fields / types** → inert, as here.
- **Enum drift** was my main worry and it does not apply. `ProductSortEnum` is the only generated enum, and the app only ever *produces* it (`ProductSort+Converter` switches over its own `SortByType.rawValue` and returns a case). So a BFF-side *addition* is inert, and a *removal* of a case you emit breaks the build. If the app ever starts *reading* an enum off a response, that inverts — Apollo wraps consumed enums as `GraphQLEnum<T>` with an `.unknown(String)` arm, and a `default:` there would swallow new cases silently. Worth remembering at that point, not now.

So the design is sound: because the schema is committed and codegen runs off the committed copy, CI and builds stay deterministic. The non-determinism is only at sync time, which is the right place for it.

The two residual gaps, both cheap:

1. **Provenance.** `sync-bff-schema.sh` does `checkout main && pull` and copies `src/schema.gql`, and nothing records *which* BFF commit produced it. When a schema change later turns out wrong, "which revision, and when" is unanswerable from this repo. Having the script emit the BFF SHA into the sync commit message would close that for about three lines of bash.
2. **The blind-inheritance point you made** — that is the one with teeth, and splitting the PR would not have fixed it. Yes please to the ticket against the bag domain; a PR thread is the wrong half-life for it.

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefinePriceView.swift
```diff
@@ -0,0 +1,56 @@
+import AccessibilityIdentifiers
+import Model
+import SharedUI
+import SwiftUI
```
khoinguyen-mindera, line 4: **🔵 Nit — `hideKeyboard()` compiles only through transitive import visibility; this breaks on Swift 6.**

This file calls `hideKeyboard()`, which is defined once in the whole package, at `Utils/Extensions/View+Extension.swift:55`. But:

- these imports are `AccessibilityIdentifiers`, `Model`, `SharedUI`, `SwiftUI` — no `Utils`;
- `ProductListing`'s target in `Package.swift` does not even list `Utils` as a dependency;
- nothing re-exports it (`@_exported import` appears only in generated `BFFGraph` files, for `ApolloAPI`).

It builds today because Swift 5 leaks *extension members* from transitively imported modules. SE-0444 `MemberImportVisibility` closes exactly that hole, so this becomes a hard error on the Swift 6 migration — in a file whose only symptom will be "why does the keyboard call not resolve".

**Fix:** add `import Utils` here, and `"Utils"` to the `ProductListing` target dependencies.

Cheap now, and it turns a future migration mystery into a no-op.

Alfie/AlfieKit/Sources/SharedUI/Components/TextInput/TextInput.swift
```diff
@@ -0,0 +1,214 @@
...
+                    }
+                )
+                .font(Font(theme.font.body.medium.uiFont))
+                .foregroundStyle(style.textColor)
+                .tint(Theme.contentContentPrimary)
+                .keyboardType(configuration.keyboardType)
+                .focused($isFocused)
+                // Applied only when supplied — an empty override would silence the field's own
+                // announcement and leave an unlabelled input.
+                .accessibilityIdentifier(configuration.accessibilityIdentifier ?? "")
```
khoinguyen-mindera: **🔵 Nit — this comment has drifted one modifier up.**

"Applied only when supplied — an empty override would silence the field's own announcement and leave an unlabelled input" describes `OptionalAccessibilityLabel` on the line below. The modifier it currently sits above, `.accessibilityIdentifier(...)`, is still applied unconditionally with an `""` fallback — which is fine (identifiers are never announced), but it makes the comment read as describing behaviour that isn't there.

Move it below `.accessibilityIdentifier(...)`, directly above `.modifier(`.

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefineViewModel.swift
```diff
@@ -0,0 +1,145 @@
...
+    /// Trailing summary on the Price row; nil when the dimension is unfiltered.
+    ///
+    /// Formatted through `CurrencyFormatter` rather than interpolating the glyph: interpolation
+    /// hardcodes symbol-before-amount, which is wrong in locales that suffix it, and `Int($0)`
+    /// traps for values the unbounded field can hold. This is also what the slider announces to
+    /// VoiceOver, so the row and the slider can no longer disagree.
+    var priceSummary: String? {
+        guard let currencyCode = priceBounds?.currencyCode, hasActiveFilters else { return nil }
+        let format = { (value: Double) in
+            CurrencyFormatter.string(amount: Decimal(value), currencyCode: currencyCode)
```
khoinguyen-mindera: **🔵 Nit — the `isFinite` guard is applied asymmetrically.**

`RangeSliderStyle.fieldText` gained `value.isFinite` as, in your words, *"defence in depth against a value arriving from anywhere else"*. This `Decimal(value)` has no such guard, and neither does the sibling in `RefinePriceView.swift:22` (`valueDescription`). `Decimal(Double.infinity)` traps — I verified it, exit 133 — which is the same fact your own `toDomainMoney` comment records.

To be clear about the severity: **this is unreachable today.** I traced every writer of `pendingMinPrice`/`pendingMaxPrice` — `init` from `appliedFilters` (itself a prior capped `filterInput`), `fieldValue` (capped at `maximumFieldValue`), `draggedValue` and `adjustedValue` (both bounded) — and none can produce a non-finite value. So this is not a bug; it is the threat model you adopted in `RangeSliderStyle` not being carried across to the two call sites that consume the same values.

Either guard these two, or drop the guard in `fieldText` and rely solely on the cap. Both are defensible; having it in exactly one of three places is the thing worth resolving.

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefinePriceView.swift
khoinguyen-mindera: Confirmed and fixed in `1e9ca51`. I verified all three legs of your claim before acting:

- `ProductListing` target dependencies are `AccessibilityIdentifiers, Core, Model, ProductDetails, Search, SharedUI` — no `Utils`.
- `hideKeyboard()` exists exactly once, `Utils/Extensions/View+Extension.swift:55`.
- This file imported no `Utils`.

So it was resolving purely through Swift 5 leaking extension members across a transitive import, exactly as you said. Added `import Utils` here and `"Utils"` to the target in `Package.swift`. Good catch — this would have surfaced during the Swift 6 migration as an unresolvable call in a file with no obvious connection to the change.

Alfie/AlfieKit/Sources/SharedUI/Components/TextInput/TextInput.swift
khoinguyen-mindera: Fixed in `1e9ca51` — moved below `.accessibilityIdentifier(...)`, directly above `.modifier(`. My sloppiness: I wrote the comment for the modifier I was adding and left it attached to the one above it.

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefineViewModel.swift
khoinguyen-mindera: Fixed in `1e9ca51`, and thank you for tracing the reachability rather than just flagging the asymmetry — that is what made the choice easy.

I went with **guard all three** rather than dropping it from `fieldText`, on this reasoning: `RangeSlider` is a public SharedUI component whose `lowerValue`/`upperValue` bindings any caller can drive, and `pendingMinPrice`/`pendingMaxPrice` are `@Published var`s writable from anywhere in the module. The cap only constrains values that arrive through `fieldValue`. So the invariant holds today because of who happens to write these, not because of anything structural — which is exactly the kind of thing that stops being true when someone adds a fourth writer.

`priceSummary`'s formatter now returns `String?` and the sides use `flatMap`, so a non-finite bound contributes nothing instead of trapping; `RefinePriceView`'s `valueDescription` guards the same way. Added `test_a_non_finite_price_does_not_crash_the_row_summary` to pin it.

khoinguyen-mindera:
## Re: `carriedFilters` — answering the open question from round 2

You asked this rather than asserting it, so it deserves a real answer rather than silence. **Keeping it, deliberately**, and here is the reasoning.

**It is not speculative generality, because the neutral behaviour is the one that preserves.** `ProductFilterInput` already has six fields — `brandNames`, `inventory`, `maxPrice`, `metafields`, `minPrice`, `productTypes`. They are part of the existing domain model; I invented nothing. `filterInput` has to construct one of these on apply, and the question is whether it constructs it from all six fields or from two. **Dropping four is the active choice that needs justifying, not preserving them.** The version without `carriedFilters` silently discards data the type is carrying — that is a behaviour I would have had to write on purpose.

**It came from a review finding, not from imagination.** Copilot flagged the absence as a defect: *"rebuilding it with only `minPrice` and `maxPrice` silently drops any existing `brandNames`, `productTypes`, `inventory`, or `metafields`."* So it is a fix, and the YAGNI test — "would I have written this unprompted?" — it passes for the wrong reason but passes.

**The failure mode is the bad kind.** A brand filter vanishing when the user touches Price is silent, plausible-looking, and would be reported as "the app forgot my filter" long after the cause. That is exactly the class `CLAUDE.md`'s simplicity rule is not trying to protect us from.

**It is also the minimal form.** Six lines of field forwarding, no abstraction. The tempting generalisation — a `ProductFilterInput.copy(minPrice:maxPrice:)` helper — *would* be speculative generality, and I deliberately did not add it. Enumerating at the one call site is the smaller thing.

**Where you are right, and it is worth recording:** the code is unreachable today, and its two tests assert behaviour no user can currently reach. I am keeping them because they pin the invariant for whoever lands the second dimension, but they are pinning a guarantee rather than covering a live path, and anyone reading them should know that. The comment on `carriedFilters` says why it exists; I have left it as the record.

If you would rather have it gone until a second dimension actually lands, say so and I will strip it plus its two tests — it is a clean removal, and your call as reviewer. But I would not make that change silently, which is why this is a reply rather than a commit.

Alfie/AlfieKit/Sources/ProductListing/UI/Refine/RefineViewModel.swift
```diff
@@ -0,0 +1,149 @@
...
+    /// than sent as an empty filter object. Price is replaced from pending state; every other
+    /// dimension is carried through as applied.
+    private var filterInput: ProductFilterInput? {
+        guard hasActiveFilters else { return nil }
+        return ProductFilterInput(
+            brandNames: carriedFilters?.brandNames,
+            inventory: carriedFilters?.inventory,
+            maxPrice: pendingMaxPrice,
+            metafields: carriedFilters?.metafields,
+            minPrice: pendingMinPrice,
```
khoinguyen-mindera: **🔵 Nit — the fourth consumer of the pending values, and the only one that leaves the process.**

`priceSummary`, `valueDescription` and `fieldText` now all guard `isFinite`. `filterInput` does not — it passes `pendingMinPrice`/`pendingMaxPrice` straight through to `BFFGraphAPI.ProductFilterInput` and into the request body.

I traced what happens if a non-finite value ever got here, because the failure mode is worse than the three display sites:

- `JSONRequest.serializationFormat` is `JSONSerializationFormat` (`Apollo/JSONRequest.swift:28`)
- which is `JSONSerialization.sortedData(withJSONObject:)` (`Apollo/JSONSerializationFormat.swift:8`)
- and `JSONSerialization` raises `NSInvalidArgumentException` — *"Invalid number value (infinite) in JSON write"* — on a non-finite number.

That is an **ObjC exception, not a Swift error**: `try` does not catch it. I verified it — SIGABRT, exit 134. So the three guarded sites degrade to an empty string, while this one would hard-crash on the network path.

**Still unreachable**, and I want to be clear I am not claiming otherwise. I re-ran the writer trace on this commit — `git grep 'pendingMinPrice ='` etc. gives exactly three: the `RangeSlider` binding (capped at `maximumFieldValue`), `init` from `appliedFilters` (itself a prior capped `filterInput`), and `removeAllFilters` (nil). None can produce a non-finite value.

I raise it only because it is the same argument you made for guarding the other three — the invariant rests on who writes, not on structure — and I miscounted the consumers when I said "three". The cheap version is folding it into the existing guard:

```swift
private var filterInput: ProductFilterInput? {
    guard hasActiveFilters else { return nil }
    let finite = { (value: Double?) in value.flatMap { $0.isFinite ? $0 : nil } }
    return ProductFilterInput(
        …
        maxPrice: finite(pendingMaxPrice),
        minPrice: finite(pendingMinPrice),
        …
    )
}
```

Your call, and this is the last time I raise it either way.

khoinguyen-mindera:
@Prathimaboggu Thanks for re-verifying the earlier criticals against the code rather than the resolved threads — that is the right instinct on a PR where the fixes and the reviews were written by the same person.

All three points taken. Fixed in `3bb9d9b`.

### The bounds latch — you were right, and the distinction is the fix

`didRequestPriceBounds` now releases in the `catch`:

```swift
} catch {
    // Only the nil result must not be re-queried — it is a legitimate answer. A throw is
    // not, so release the latch: otherwise one transient failure hides the Price row for
    // the life of the screen, with no path back to it.
    didRequestPriceBounds = false
    dependencies.log.error("Error fetching category price range: \(error)")
}
```

I kept the pre-await set, as you suggested — it still dedupes concurrent `viewDidAppear` calls, and that was the reason it sat there in the first place. Your reading of my own doc comment is the part I had wrong: "latch on asked, not on got a result" is sound reasoning *about the nil case*, and I wrote a guard that could not tell the two apart. The comment now says which case it means.

### `didApplyFilters` doesn't clear `refreshError`

Fixed, one line next to the `pagination = nil` reset:

```swift
// A failed pull-to-refresh describes the previous result set; leaving its Snackbar up over
// a freshly filtered listing reads as the filter having failed.
refreshError = nil
```

### The `filterInput` non-finite nit

Taken. `filterInput` now runs both values through the same `finite` closure, so all four consumers of the pending values share one threat model rather than three of four being armed. You and I both traced it as unreachable and it is; the argument that won for the other three consumers is uniformity, not reachability, and it applies here identically.

### Tests

Two added, and I checked they are not vacuous — I stashed the view-model fix and re-ran the suite against the unfixed code:

- `test_a_thrown_bounds_fetch_is_retried_on_the_next_appearance` — sits directly after `test_a_category_without_bounds_is_not_requeried_on_every_appearance`, so the pair reads as the two halves of the latch rule, which is what you asked for. Against the unfixed view model it times out on the `$priceBounds` expectation and reports 1 fetch where 2 are expected.
- `test_applying_filters_clears_a_stale_refresh_error` — fails with `XCTAssertNil failed: "serverError"` before the fix.

No test for the `isFinite` guard: no writer can produce the value, which is the same reason we both ranked it a nit.

### Verification

`./Alfie/scripts/verify.sh` → **✅ FULL VERIFICATION PASSED** (build + unit + 10 integration tests against a real local BFF).

One thing worth knowing if you run verification from a git worktree: the script resolves the BFF as `../Alfie-BFF` relative to the repo root, which from `.claude/worktrees/…` points nowhere, so the integration step fails on a missing repo rather than on anything real. `ALFIE_BFF_PATH` overrides it. Not part of this PR, but I will raise it separately.

Prathimaboggu (approved):

PR #104, opened 2026-08-03
ALFMOB-469: Modernize Shop category menu + BFF menu-query fix

khoinguyen-mindera:
## Ticket
[ALFMOB-469](https://mindera.atlassian.net/browse/ALFMOB-469)

## Summary

**Category rows — modern Figma restyle**
- Migrate `CategoriesView` rows off legacy `Primitives.*` to semantic tokens: `Theme.contentContentPrimary` / `Theme.contentContentPrimaryDisabled` for text, `theme.spacing.space0` / `space200` for spacing.
- Match the Figma menu-item spec: row height 56 → 48, remove the per-row `ThemedDivider`, grow the chevron 16 → 24 (via `Sizing.iconsIconMedium`).
- Root (level 1) menu uses `body.large` (18/28); drill-down levels use `body.medium` (16/24), selected via a new `CategoriesViewModelProtocol.isRoot`.
- Delete the dead `segmentedControlHeight` constant.

**Shop toolbar → search-bar entry**
- Replace the Shop root toolbar (title + wishlist/account icons) with a `.soft` `ThemedSearchBarView` header that presents the existing Search flow on tap (`presentSearch()`); wishlist/account remain reachable via the bottom tabs.
- Make the search entry VoiceOver-operable — wrap it in a `Button` with the a11y id/label, visual bar `.accessibilityHidden(true).allowsHitTesting(false)` (applied to both Shop and Home for consistency).
- Remove the now-dead `CategorySelector+Toolbar.swift`, the orphaned `isWishlistEnabled` from the flow VM/protocol, and unused search a11y ids.

**BFF `menu` query fix** (BFF ticket AF-95)
- The BFF renamed the query field `mainMenu` → `menu` (arg + `Menu` shape unchanged), so the Shop menu failed against the latest BFF. Update `MainMenuQuery` to request `menu(handle:)`, apply the `menu` field rename to the committed schema (**menu-only edit — not a full BFF re-sync**), regenerate the Apollo API/mocks, and update `BFFClientService` / `MainMenu+Converter` / tests.
  - Note: the committed `schema.graphqls` still lags BFF `main` in unrelated areas (e.g. the `updateCart` mutation added 2026-07-27); a full `sync-bff-schema.sh` is left as a separate follow-up to keep this PR menu-scoped.

## Testing
- `./Alfie/scripts/verify.sh --skip-integration` — build + unit tests pass.
- Ran in the simulator against the latest `main` BFF: Shop categories load (Women/Men/Kids/Beauty), drill-down works (Women → Bags/Clothes/Shoes/Sports), and a leaf category opens its PLP with real products.

## Screenshot
Simulator Screenshot - iPhone 17 Pro Max - 2026-08-04 at 11 26 41
Simulator Screenshot - iPhone 17 Pro Max - 2026-08-04 at 11 26 43

[ALFMOB-469]: https://mindera.atlassian.net/browse/ALFMOB-469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Alfie/AlfieKit/Sources/CategorySelector/Navigation/CategorySelectorRoute+Destination.swift
```diff
@@ -25,13 +23,9 @@ public extension CategorySelectorRoute {
...
-            ShopView(
-                isRoot: isRoot,
-                isWishlistEnabled: isWishlistEnabled,
-                categoriesViewModel: categoriesViewModel()
-            ) {
-                navigate($0)
-            }
+            // Unreachable: `navigate(.categorySelector)` pops to root rather than pushing. The live
+            // root ShopView is built directly in CategorySelectorFlowView.body.
+            EmptyView()
```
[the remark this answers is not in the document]
khoinguyen-mindera: Good call — addressed in dce2ecd. The `.categorySelector` case now renders `CategoriesView(viewModel: categoriesViewModel())` as a graceful fallback instead of `EmptyView()`. It's still unreachable in normal flow (`navigate(.categorySelector)` pops to root), but this avoids a blank screen if it's ever pushed via a deep link or future call site.

Prathimaboggu (commented):
@khoinguyen-mindera

**Static review only** — Command Line Tools box, no iOS SDK, so no build/tests/lint on my side. What I did verify: the committed GraphQL schema against `Alfie-BFF@origin/main`, the Figma `Menu Sheet – Level 1` node (1:540) in dev mode, and call-site traces for the changed signatures. I couldn't run the Home snapshot suite, so that one's unconfirmed — statically it looks unaffected.

Copilot's `EmptyView()` question checks out, for the record: `CategorySelectorFlowViewModel.navigate` maps `.categorySelector` to `popToRoot()` (`CategorySelectorFlowViewModel.swift:248-256`) and never appends it to `path`, and the only external entry — the `.shop` deep link in `AppFeatureViewModel.swift:221` — goes through that same method.

Nothing blocking. Two things I'd fix before merge:

**1. `BFFGraph/CodeGen/Schema/schema.graphqls` — the schema was hand-edited, not re-synced.**

The description says "re-sync the committed schema", but diffing against `Alfie-BFF@origin/main:src/schema.gql` shows it's still missing `updateCart(input: UpdateCartInput!): Cart!` from `type Mutation`, plus `input UpdateCartInput` and `input UpdateCartLineInput`. Those landed in the BFF on 2026-07-27 (`349b595 feat(AF-88) implement updateCart for Shopify`), a week before this PR's sync commit; the single `mainMenu` → `menu` line is the only other difference. Either re-run the sync so the next person doesn't inherit a surprise diff, or reword the description to "updated the menu field". (The rename itself is real — `f1dfab6 AF-95` is on BFF `main` and the `type Query` block is now byte-identical.)

**2. `ShopView.swift:29-42` — the search bar is 8pt shorter than the Figma, so the whole menu sits 8pt high.**

Figma `Menu Sheet – Level 1` (1:540) places the Search instance (17:15275) at 343×**40**, 4pt below the status bar, with 20pt down to the first menu row — 64pt from status bar to first row. The implementation uses `ThemedSearchBarView(theme: .soft)`, whose `searchBarHeight` is **32** (`ThemedSearchBarView.swift:95-104`), plus `.padding(.top, space100)` = 8 and `.padding(.bottom, space200)` = 16, landing the first row at 56pt. Neither `.soft` (32) nor `.softLarge` (44) yields 40, so this needs a design/DS call rather than a token swap — and since it matches the already-shipped Home header exactly, the answer may be that the Figma node is the stale one. Everything else on the rows matches Figma exactly: 48pt height, contiguous rows with no dividers, 16pt horizontal inset, `#111111`.

khoinguyen-mindera:
@Prathimaboggu thanks for the thorough static review 🙏

**1. schema hand-edit vs re-sync** — fair, reworded the description: this was a **menu-only** edit (`mainMenu` → `menu`), not a full `sync-bff-schema.sh`. The `updateCart`/`UpdateCartInput` drift you spotted pre-dates this PR (BFF `349b595`, 2026-07-27) and is unrelated to the menu change, so I've kept it out of scope and noted a full-schema-sync as a separate follow-up rather than pulling an unused `updateCart` mutation + regenerated codegen into a menu PR. Happy to file that ticket.

**2. search bar 32pt vs Figma 40pt** — good catch. It's a deliberate DS match, not a token miss: `.soft` is 32pt and intentionally mirrors the already-shipped Home search header; no DS token yields 40 (`.soft`=32, `.softLarge`=44). Since 40 isn't expressible with current tokens and the Figma node may be stale, I'd rather not hardcode 40 or diverge from Home unilaterally — flagging for a design/DS decision. If design confirms 40, we add/adjust a token and apply it to both Home and Shop together.

Everything else (rows: 48pt, no dividers, 16pt inset, `#111111`) matches Figma. Copilot's `EmptyView()` point is addressed in dce2ecd (renders the categories list as a graceful fallback).

Alfie/AlfieKit/Sources/CategorySelector/UI/ShopView.swift
```diff
@@ -40,12 +52,10 @@ struct ShopView<CategoriesViewModel: CategoriesViewModelProtocol>: View {
...
     ShopView(
-        isRoot: true,
-        isWishlistEnabled: true,
         categoriesViewModel: MockCategoriesViewModel(
             state: .success(.init(categories: [])),
             categories: NavigationItem.fixtures
         )
-    ) { _ in }
+    ) {}
 }
```
chaitanya-mindera, line 48: Minor, non-blocking

ShopView.swift:15-31 and HomeView.swift:20-36 now carry near-identical Button-wrapping-a-disabled-ThemedSearchBarView blocks (same .allowsHitTesting(false)/.accessibilityHidden(true)/.buttonStyle(.plain) pattern, only HomeView's has .matchedGeometryEffect). Not asking for an extraction in this PR, but if a third search entry point shows up, this should become a shared component rather than a third copy — the two are already one accidental edit away from drifting (e.g. only Home got the geometry effect).

khoinguyen-mindera: Good call — went ahead and extracted it now while both call sites are fresh (f03b22f). New `SearchBarEntryButton` in SharedUI (`Theme/Components/SearchBar/`) owns the Button + `.allowsHitTesting(false)`/`.accessibilityHidden(true)`/`.buttonStyle(.plain)` wrapping and the a11y label/id; Shop and Home now both call it. Callers keep only their own outer layout — Home still applies `.matchedGeometryEffect` on the component (its frame matches the inner bar, and that namespace is Home-local anyway). No more drift risk between the two.

hoangnhatdrk (commented):
Reviewed the latest state. The prior threads look resolved: the `.categorySelector` → `CategoriesView` fallback (dce2ecd), the `SearchBarEntryButton` extraction (f03b22f), the schema-description reword, and the 32pt-vs-Figma-40pt height flagged for a design/DS call.

Two things I checked independently — both come back clean:

- **Removing `AccessibilityID.Home.searchBackButton` is safe.** The id has no references outside `HomeView`/`AccessibilityID.swift` (no UI test uses it), so nothing breaks.
- **Dropping `isWishlistEnabled` from this flow does not ungate wishlist.** `AppFeatureViewModel` still gates it via `isFeatureEnabled(.wishlist)` at the tab level, so the config flag is still honoured — the removed check was only gating the old Shop toolbar icon.

No blockers from me. Remaining open items are both already-known and non-code: the search-bar height (deliberate match to the shipped Home header — needs the design/DS decision, not a code change) and the full `sync-bff-schema.sh` follow-up ticket. LGTM once the height question is settled with design.

hoangnhatdrk (approved):
Approving — see my review notes above. Nothing blocking; the remaining items are non-code (design/DS decision on the search-bar height for #104) or non-blocking nits.

PR #107, opened 2026-08-10
ALFMOB-441: PDP foundation — single-scroll layout and design-token migration

khoinguyen-mindera:
## Ticket
[ALFMOB-441](https://mindera.atlassian.net/browse/ALFMOB-441)

## Summary
- Product Details renders as one continuous scroll; the bottom sheet, its detents and the content measurement that fed them are removed
- _Add to Bag_ and the wishlist button now sit directly beneath the price and scroll with the page
- The iPad and pre-iOS-16.4 paths are collapsed into the single scroll; `isIpad` is retained only for horizontal padding
- Product Details migrated to design tokens — 53 legacy primitive references replaced, one documented exception (`neutrals600` has no `Theme.*` alias, see `.scratch/pdp-modern-design/token-requests.md` G3)
- New `ProductDetailsLayoutRules` holds the size-layout, colour-layout, low-stock and swatch-appearance rules in one testable place (19 unit tests, incl. nil and zero stock)
- `Typography.Body.mediumBold` exposed via `TypographyBody` and bound to the price, replacing a hand-rolled weight override
- `ThemedButton` gains a `cornerRadius` parameter defaulting to `Sizing.radiusSoft`; the PDP call-to-action and wishlist button pass `0` for square corners, leaving the other 38 call sites unchanged
- `selectedColourName` and `productReference` added to the view-model protocol and mock; product fixtures now carry real brand, colour and reference values instead of empty strings and UUIDs
- First snapshot coverage for Product Details — default and error states
- Design tokens refreshed from upstream (`Typography.Body.mediumBold`, `colours-brand-*` ramp, `spacing44`, `contentContentSecondary`)

## Screenshot

[ALFMOB-441]: https://mindera.atlassian.net/browse/ALFMOB-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hoangnhatdrk (commented):
The PDP-scoped work is clean and genuinely well-tested — `ProductDetailsLayoutRules` is a nice pure extraction, the `canShow*` rewrites are behaviour-preserving (`!= .summaryOnly` ≡ `count > 1`; `== .verticalList` ≡ `count > 6`), the sheet/detent removal is a real simplification, and the price → `mediumBold` migration preserves the effective weight.

Before approving, a few things — mostly about blast radius beyond PDP:

**1. App-wide changes bundled into a PDP PR.** `buttonPrimaryBackgroundPrimaryDefault` / `…StrokePrimaryDefault` remap `neutrals800 → brand500` (~#111111), which repaints **every** primary button in the app; and the `Product`/`Brand` fixture defaults change from empty strings / UUIDs to real values, which ripples through every test and snapshot that builds a `Product.fixture`. This PR only *adds* PDP snapshots — please confirm the full suite (Home / Splash / PLP / etc. snapshot baselines included) has been re-run/re-recorded and non-PDP screens were checked. The spec committed in this very PR says the token refresh "ships as its own PR because its generated diff is app-wide" — worth splitting the token + fixture changes out so the PDP review isn't coupled to an app-wide visual change.

**2. `manifest.json` now points at `theme.new-brand-theme.tokens.json`, which isn't in the repo** (the old `theme.selfridges.tokens.json` isn't either). The generator appears to tolerate a missing theme entry — selfridges was already absent — so this is probably benign, but since it's a hand-edited generated manifest, please confirm regeneration still works and this is an intended rename rather than a stray edit.

**3. Copilot's flaky-test point is still open** — `test_withSize_preservesWeight` asserts an exact `CGFloat` weight-trait equality; worth switching to an accuracy tolerance, consistent with the sibling assertions in that file.

**4. Minor.** `ProductDetailsLayoutRules.stockMessage(...)` and `sizeSwatchAppearance(...)` are fully tested but not consumed yet (deferred to the size-selector slice, per the in-code comment) — fine as staged code, just flagging it isn't wired in this slice. Also, the old `neutrals900` maps to two different tokens across the two image `failure:` branches (`surfaceBackgroundInvertedPrimary` vs `contentContentPrimary`) — pick one unless there's a reason.

Happy to approve once (1) is confirmed green and (2) is clarified.

khoinguyen-mindera:
Thanks — two real defects found, both fixed in b5970f2. Evidence for the rest below.

**4b — inconsistent `failure:` token. Correct, my mistake.** The two carousels are indented differently, so my scripted replace matched only `mediaCarousel` and the fullscreen one fell through to the bulk pass. Both now use `Theme.surfaceBackgroundInvertedPrimary` — a dark filled surface standing in for a missing image, matching the swatch fallback in `ProductDetailsViewModel`.

**3 / Copilot — flaky weight assertion. Fixed.** Now unwraps and compares with `accuracy: 0.001`.

**1 — app-wide blast radius. Confirmed green; splitting is your call.**

The button remap is provably a no-op rather than a repaint — both primitives are byte-identical in the generated output:

```swift
public static let brand500    = Color(.sRGB, red: 0.066667, green: 0.066667, blue: 0.066667, opacity: 1.0)
public static let neutrals800 = Color(.sRGB, red: 0.066667, green: 0.066667, blue: 0.066667, opacity: 1.0)
```

All three snapshot suites run in the full unit suite and pass against **unchanged** baselines — only the two new PDP ones were recorded:

```
SplashViewSnapshotTests test_splashView                      passed
HomeViewSnapshotTests   test_homeView_withHeroBanners        passed
HomeViewSnapshotTests   test_homeView_withoutHeroBanners     passed
ProductDetailsViewSnapshotTests (2 new)                      passed
```

Those are the only three suites in the repo — PLP has no snapshot coverage yet. The fixture change also broke nothing: no existing test depended on the empty-string defaults. `id`/`sku` deliberately stay UUIDs because `resolvedSelectedVariant` matches on SKU uniqueness.

On splitting: agreed the spec asks for it, and the refresh is already isolated in `71da217` (`chore: refresh design tokens from upstream and regenerate`), so it cherry-picks cleanly onto its own PR. Happy to do that if you'd prefer — flagging only that ticket 01 depends on it for `Typography.Body.mediumBold`, so this PR would need to land after it.

**2 — `manifest.json`. Not a hand edit, and pre-existing.**

`manifest.json` is copied verbatim by `pull-design-tokens.sh`; the only change is upstream's `selfridges` → `new-brand-theme` rename. The referenced file is absent **by design** — the script pulls only the iOS subset (`TOKEN_FILES` lists `theme.alfie-theme.tokens.json` and no second brand theme). `theme.selfridges.tokens.json` was equally absent on `main`, so nothing changed but the name.

The generator pins the theme collection to `alfie-theme` and has a regression test for exactly this: *"theme is pinned to alfie-theme, skipping the file-less selfridges mode (regression: dual-mode theme used to load every mode's file)"*. Regeneration was run three times against upstream `main` @ `a3b2b88` with its 42 generator tests passing each time.

**4a — staged-but-unconsumed rules.** Correct and intended: AC #8 asks for the rules type as groundwork for the size-selector slice. `stockMessage` also carries an `// ASSUMPTION:` on its threshold — design hasn't confirmed N.

Two caveats on this PR that are worth stating plainly:
- **Integration tests were not verified locally.** `Alfie-BFF/.env` has no `PLATFORM`, and booting with `PLATFORM=mock` gave `noProduct` / a 400 on header-nav. The branch touches zero networking code and `test_getProduct_unknownHandle` passed, so CI should be the gate here.
- **The CTA reachability check the ticket asks for** (small device, long-description product) is still outstanding — it needs a working BFF.

khoinguyen-mindera:
Update on the three remaining points.

**1 — token refresh split out: #108.**

`chore/design-token-refresh` branches off `main` and carries only the 8 generated/pulled token files. `./Alfie/scripts/verify.sh --skip-integration` passes on that branch **alone**, with all three snapshot suites green against unchanged baselines — which is the cleanest available proof that the app-wide token diff repaints nothing.

Please merge #108 first. This PR depends on it for `Typography.Body.mediumBold`. I have deliberately **left the token commits on this branch** so it keeps building in the meantime; they will collapse to nothing once #108 lands, and I will rebase then.

**2 — `manifest.json`: nothing to change.** Evidence in my previous comment — the file is copied verbatim by `pull-design-tokens.sh`, the absent theme file is the intended iOS subset, and the generator pins to `alfie-theme` with a regression test naming that exact case. Same explanation now also written into #108's description.

**4 — the two verification gaps are tracked, both deferred by agreement:**

- *Integration tests.* Alfie-BFF now requires a `PLATFORM` env var with no default, and `run-integration-tests.sh` does not set one. `PLATFORM=mock` boots the server but serves no seed data. Fix belongs in the verify script, in its own PR.
- *Small-device CTA reachability.* Will be covered by adding a small-device snapshot case rather than by manual testing, also in a later PR.

Neither is caused by this change: the branch touches no networking code, and `test_getProduct_unknownHandle` plus the whole ProductList suite pass against the same server.

khoinguyen-mindera:
@hoangnhatdrk ready for another look when you have a moment.

Since your review:

- **4b — you were right**, the two `failure:` branches had drifted to different tokens. Both now use `Theme.surfaceBackgroundInvertedPrimary` (`b5970f2`).
- **3 / Copilot** — the weight assertion now unwraps and compares with `accuracy: 0.001`.
- **1 — token refresh split out into #108**, branched off `main`. It passes `verify.sh` standing alone, with all three snapshot suites green against unchanged baselines. **Please merge #108 before this one** — ticket 01 needs `Typography.Body.mediumBold` from it. The token commits stay here for now so this branch keeps building; I will rebase once #108 lands.
- **2 — `manifest.json`** needs no change; it is copied verbatim by `pull-design-tokens.sh` and the absent theme file is the intended iOS subset. Detail in my earlier comment and in #108.
- **4a** — the unconsumed rules are intentional groundwork for the size-selector slice.

The two verification gaps (BFF `PLATFORM` for integration tests, small-device CTA check) are deferred to their own PRs by agreement, and neither is caused by this change.

hoangnhatdrk (approved):
Approving — all four points from my previous review are addressed and I verified each in the diff:

- **4b** — both image `failure:` branches now use `Theme.surfaceBackgroundInvertedPrimary`, matching the swatch fallback. ✅
- **3 / Copilot** — the weight assertion now unwraps and compares with `accuracy: 0.001`. ✅
- **1** — the app-wide token diff is split into #108; the button remap is a verified no-op (`brand500` == `neutrals800` == `#111111`) and the Splash/Home snapshots pass against unchanged baselines. ✅
- **2** — `manifest.json` is copied verbatim by `pull-design-tokens.sh`, the absent theme file is the intended iOS subset, and the generator pins `alfie-theme` with a regression test. ✅
- **4a** — the unconsumed layout rules are intentional groundwork for the size-selector slice. ✅

Two notes, neither blocking:
- **Merge order:** #108 must land first — this branch still carries the token commits, and it depends on #108 for `Typography.Body.mediumBold`. Please rebase after #108 merges.
- The token snapshots on the two branches have slightly diverged — `contentContentSecondary` is `neutrals800` here vs `neutrals700` on #108 — so the rebase won't be a pure no-op. It's harmless because nothing consumes `contentContentSecondary` in this PR (zero call sites), so no snapshot moves.

chaitanya-mindera (approved):

